PR #4, opened 2022-01-18
Level 5 Functionality

dandaandaaaaaan:
Added error handling and updated test cases for following cases

- Invalid command
- Todo task with no title
- Deadline task with no title
- Deadline task with no deadline
- Event task with no title
- Event task with no event timing

No error is handled for repeated marking/unmarking (eg. already marked event is marked again) as it is presumed that it will stay marked. This is intended behaviour


dandaandaaaaaan:
My bad, accidentally opened PR to wrong branch

PR #5, opened 2022-01-19
[Sim Si Leng] iP

pnutzz-0207:
## SONA

> "The future depends on what you do today." - Mahatma Gandhi 

Hi! My name is sona, a ~~good~~ SUPER AMAZING task manager. I am:

- text based
- easy to learn

Being a personal assistant that helps to manage your task, I do more than just recording tasks!

- [X] Categorize your task into different types: to-do, deadline, event
- [X] mark/unmark your task
- [X] Lookup for tasks on a specific day
- [X] Search tasks by keyword

Most importantly, I am is totally **free**!

It is simple have me in your computer! Simply...

1. Click [here](https://github.com/pnutzz-0207/ip/releases/tag/A-Jar) to download ⏬
2. double-click it 🖱️
3. add your tasks...and TADA! She will manage your task for you ✨

If you want to be a Java programmer, you can use me to practice too! Here is the main method:
```ruby
        new Duke("src/main/data/duke.txt").run();
}
```


src/main/java/duke/sonautil/Ui.java
```diff
@@ -0,0 +1,325 @@
...
+        return "Hi! You can call me Sona, your AMAZING task manager ><! How can I help you?";
+    }
+
+    /**
+     *  returns string after user types "bye"
+     *
+     * @return message to user
+     */
+    public static String goodbyeMessage() {
+        return "Bye-bye :')  Hope to see you soon!";
```
AAlghrairy, line 46: Perhaps rewrite the method name to be a verb (e.g. returnGoodbyeMessage)?

src/main/java/duke/Main.java
```diff
@@ -0,0 +1,37 @@
...
+
+/**
+ * A GUI for Duke using FXML.
+ */
+public class Main extends Application {
+
+    private final Duke duke = new Duke("src/main/data/duke.txt");
+
+    @Override
+    public void start(Stage stage) {
```
AAlghrairy, line 21: Perhaps create a header comment for this method (since it is a public method)?

src/main/java/duke/sonautil/Parser.java
```diff
@@ -0,0 +1,236 @@
...
+     * arr[1]: string representation of Duke.task index (if applicable, null otherwise)
+     * arr[2] : string representation of time (if applicable, null otherwise)
+     * arr[3]: "true" if time is entered by user, "false" if not entered; null if not applicable
+     *
+     * @param userMessage the command entered by user
+     * @return details of user's command
+     * @throws DukeException If command is invalid / not understood
+     * @throws DateTimeParseException If date/time entered is invalid
+     */
+    public String[] messageProcess(String userMessage) throws DukeException, DateTimeParseException {
```
AAlghrairy: Perhaps rewrite the method name to be in the form of a verb (e.g. "processMessage")? I noticed this issue in a few other places.

src/main/java/duke/sonautil/Parser.java
```diff
@@ -0,0 +1,236 @@
...
+            command[0] = "find";
+            command[1] = keyword;
+
+        } else {
+            command[0] = "unknown";
+            throw new DukeException(Ui.userUnknownInputMessage(userMessage));
+        }
+
+        return command;
+    }
```
AAlghrairy: Consider using enums here to make the processing of the user input simpler (e.g. create an Enum for task types and pass the type here for easier processing, rather than repeated calls to "startsWith()". Also, any reason why you did not use a switch statement here rather than repeated using "else if"?

src/main/java/duke/sonautil/Storage.java
```diff
@@ -0,0 +1,226 @@
...
+        default:
+        }
+    }
+
+    /**
+     * deletes the Duke.task from local file by creating a temp file
+     * and copy every Duke.task except the one to be deleted
+     *
+     * @param taskToDelete the index of Duke.task to delete
+     */
```
AAlghrairy: The "Duke." before each reference to tasks and other Duke related functions is most likely due to a refactor bug (i.e. from changing the name of a package, or moving a file to a new package). I noticed the same issue elsewhere in the code documentation. Consider deleting each of these modifications by the program in the interest of clarity.

```suggestion
    /**
     * Deletes the task from local file by creating a temp file
     * and copy every task except the one to be deleted
     *
     * @param taskToDelete the index of task to delete
     */

File: src/main/java/duke/sonautil/Storage.java [hunk 6]
```diff
@@ -0,0 +1,226 @@
...
+        } catch (IOException e) {
+            e.printStackTrace();
+        }
+
+        File oldFile = new File(oldFileName);
+        oldFile.delete();
+
+        File newFile = new File(tempFileName);
+        newFile.renameTo(oldFile);
+    }
```
AAlghrairy: I like your implementation of deleting tasks from the file, rather than rewriting the entire file. 👍

AAlghrairy (commented):
Looks Good 👍.

Just a few coding standard related issues and a few suggestions.

src/main/java/duke/Main.java
pnutzz-0207: Yep! Thanks for the reminder!

src/main/java/duke/sonautil/Parser.java
pnutzz-0207: Thanks for the suggestion! it makes sense :)

PR #12, opened 2022-01-20
[Toh Zhan Qing] iP

tzhan98:
#### Update on Duke java project
- Level 1, 2, 3, 4, 5, 6, 7, 8, 9
- A-TextUiTesting
- A-JavaDoc
- CodingStandard

> "Your mind is for having ideas, not holding them.” – David Allen [source](https://github.com/tzhan98/ip)

- text-based
- easy to learn
- ~~FAST~~ SUPER FAST to use

All you need to do is,

1. download it from [here](https://github.com/tzhan98/ip).
2. double-click it.
3. add your tasks.
4. let it manage your tasks for you 😉 👍 

And it is **_FREE!_**

Features:

- [ ] Managing tasks
- [ ] Reminders (coming soon)

```
public class Main {
    }
}


```


src/main/java/duke/Deadline.java
```diff
@@ -0,0 +1,31 @@
+package duke;
+
+/**
+ * This is a type of duke.Task: deadline
+ *
+ */
+public class Deadline extends Task{
```
FestiveCat: Should there be no spaces between `Task` and `{`? (I noticed this in other places too)

src/main/java/duke/Deadline.java
```diff
@@ -0,0 +1,31 @@
...
+ */
+public class Deadline extends Task{
+
+    /**
+     * Constructor of deadline.
+     * Have to strip/process the input first and remove 'deadline'
+     *
+     * @param input descriptor of task
+     * @param number task number
+     * @param deadline deadline to complete task
```
FestiveCat: hmmm, there seems to be a missing `@param reading`, is this intentional?

src/main/java/duke/TaskList.java
```diff
@@ -0,0 +1,88 @@
+package duke;
+
+import java.util.ArrayList;
+
+/**
+ * This is the class that stores the task in a list format
+ * and provide methods to modify the list
+ */
+
```
FestiveCat: What's your rationale for including this blank line here?

src/main/java/duke/Ui.java
```diff
@@ -0,0 +1,133 @@
...
+ */
+
+public class Ui {
+
+    public Ui(){
+        startup();
+    }
+
+    /**
+     * Main program that Duke runs on, will loop till input is 'bye'
```
FestiveCat: Hmm shouldn't the Javadoc comment be better if it started with a singular verb? For example:
```ruby
/**
 * Loops user input until a "bye" is received. This is the main program that Dukes runs on.
 */
```

src/main/java/duke/DukeExceptions.java
```diff
@@ -0,0 +1,16 @@
...
+/**
+ * Placeholder Exception
+ */
+public class DukeExceptions extends Exception{
+}
+
+/**
+ * Exception to be thrown if descriptor is empty
+ */
+class EmptyDescriptorExceptions extends Exception{
```
FestiveCat: Any reason why you have two classes in the same Exceptions File?

src/main/java/duke/TaskList.java
```diff
@@ -0,0 +1,88 @@
...
+public class TaskList {
+    public ArrayList<Task> tasklist;
+
+    public TaskList(){
+        tasklist = new ArrayList<Task>();
+    }
+
+
+    public void markTaskNum(int taskNum, String check){
```
FestiveCat: Do you think there's a better way to name the String parameter `check` ?

src/main/java/duke/TaskList.java
```diff
@@ -0,0 +1,88 @@
...
+        }
+    }
+
+    /**
+     * Deletes a specified index (starts from 1) from the input ArrayList and shifts
+     * all subsequent task numbers accordingly by +1.
+     *
+     * @param num index (starts from 1) to delete
+     */
+    public void deleter(int num){
```
FestiveCat: Is it intentional to name this method `deleter`? Or did you mean to write `delete` or `deleteTask`?

src/test/java/duke/DeadlineTest.java
```diff
@@ -0,0 +1,15 @@
+package duke;
+
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+public class DeadlineTest {
+    @Test
+    public void toStringTest(){
```
FestiveCat: I think it would be good if you included the test parameters and the expected outcome. For example:
```ruby
public void toStringTest_deadlineTask_success() {
```

src/main/java/duke/Task.java
```diff
@@ -0,0 +1,127 @@
...
+import java.time.format.DateTimeFormatter;
+import java.time.format.DateTimeParseException;
+
+
+/**
+ * Class task that duke.Duke creates
+ */
+public class Task {
+
+    String time = "";
```
FestiveCat: What's your reason for leaving these as `package-private` instead of `private`?

FestiveCat (commented):
Might want to be a little more careful in following the coding standard! Logic mostly looks fine though, good job!

src/main/java/duke/DukeExceptions.java
tzhan98: I was thinking so that It would be easier to refer to since all of my custom exception classes will be just a few lines. Is it confusing?

FestiveCat: Hmm I'm of the opinion that each file should only contain one class (other than nested classes), but if its just a few lines I don't see any harm done too

src/main/java/duke/Deadline.java
tzhan98: Not intended. Done my javaDocs then made some edits to function without updating it.

src/main/java/duke/TaskList.java
tzhan98: There was no rationale, just a mistake 😆

src/main/java/duke/TaskList.java
tzhan98: Changed it to `isTrue`

src/test/java/duke/DeadlineTest.java
tzhan98: Made necessary changes

src/main/java/duke/Task.java
tzhan98: If you are talking about the fields in class `Task` such as `time`, it is because other classes in this package will read it. Would making it private and having a `getter()` be better?

FestiveCat: Hmm think in your case that might be a good idea? Since the `Task.java` is inside the package duke and not in a special package like duke.tasks

src/main/java/duke/DukeExceptions.java
```diff
@@ -0,0 +1,16 @@
+package duke;
+
+/**
+ * Placeholder Exception
+ */
+public class DukeExceptions extends Exception{
```
SimJM: DukeExceptions does not have any body, can be removed if you are not intending to work on it.

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,112 @@
...
+            } else {
+                s = Output.printEmptyDescriptionException();
+            }
+        } else if (input.startsWith("find ")){
+            s = taskList.findTask(input.substring(6));
+        } else {
+            s = Output.printWhatDoesThatMean();
+        }
+        return s;
+    }
```
SimJM: You can consider using a switch statement over here

src/main/java/duke/Task.java
```diff
@@ -0,0 +1,138 @@
...
+
+import java.io.IOException;
+import java.time.LocalDateTime;
+import java.time.format.DateTimeFormatter;
+import java.time.format.DateTimeParseException;
+
+/**
+ * Class task that duke.Duke creates
+ */
+public class Task {
```
SimJM: Looks neat!

src/main/java/duke/TaskList.java
```diff
@@ -0,0 +1,73 @@
...
+            s += Output.printMarkTaskNotDone(curr);
+        }
+        Storage.writeAllToFile(this);
+        return s;
+    }
+
+    public String findTask(String keyword) {
+        TaskList matchTasks = new TaskList();
+        for(int i = 0; i < Task.totalTask; i++) {
+            if (this.tasklist.get(i).name.contains(keyword) || this.tasklist.get(i).time.contains(keyword)) {
```
SimJM: Code can be neater by using a variable for "this.tasklist.get(i)".

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,23 @@
+package duke;
+
+/**
+ * This is the main duke.Duke program that will be able to process a duke.Task of 3 types: todo, deadline and task
+ * duke.Duke is able to list, delete and mark/unmark tasks as done/undone.
+ *
+ * @author Toh Zhan Qing
+ */
+public class Duke {
```
SimJM: Overall code is neat and coding standards are more or less there. Nice work!

PR #14, opened 2022-01-20
[Lin Hui Xin Tiffany] iP

Tiffanylin21:
# Duke
> This is a description of Duke, the CLI task manager

Duke is your pocket task manager, it is:
- text-based
- ~fast~ *super* fast
- easy to learn
All you need to do is: 
1. Download it from [here](https://github.com/Tiffanylin21/ip.git)
2. Add your tasks
3. Let Duke manage your tasks for you!
 And it's **free**!

Features:
- [x] Manage Events, Deadlines and Todos
- [x] Mark tasks as done when complete :tada:
- [ ] Simple and user-friendly GUI for a hassle-free experience (Coming soon)

If you would like to specify a folder and file for storage, please modify the `main` method which looks like this:
```java
    String dirPath = "./src/main/data";
    String filePath = "./src/main/data/duke.txt";
    Duke duke = new Duke(dirPath, filePath);
    duke.run();
}
```


src/main/java/duke/helpers/TaskList.java
```diff
@@ -0,0 +1,222 @@
...
+
+import java.util.ArrayList;
+
+/**
+ * Represents an array of Tasks which the user can perform operations on including adding,
+ * deleting, marking and getting specific Tasks.
+ */
+public class TaskList {
+
+    private static ArrayList<Task> tasks;
```
snss231, line 20: Curious to hear why you chose to set tasks as a class-level member, in my opinion it would make sense for each instance of TaskList to have it's own ArrayList<Task>

src/main/java/duke/exceptions/InvalidIndexException.java
```diff
@@ -0,0 +1,7 @@
+package duke.exceptions;
+
+public class InvalidIndexException extends DukeException {
```
snss231, line 6: Cool that you created subclasses for DukeException :)

src/main/java/duke/helpers/Parser.java
```diff
@@ -0,0 +1,97 @@
...
+
+    /**
+     * Determines the type of commands given in the user input.
+     *
+     * @param s user input.
+     * @param command type of command to be checked against.
+     * @return boolean that indicates if the user gave the specified command.
+     * @throws DukeException If deadline and event commands are missing date and/or time.
+     */
+    static boolean isCommand(String s, Command command) throws DukeException {
```
snss231, line 24: I really like the idea of having static methods for Parser actually, then we can just use Parser like a library of static methods since we don't really need to instantiate multiple Parser objects each with their own state. I might make that change for my own implementation

snss231 (commented):
Really nice polished implementation 😄 , just shared some thoughts about static members above

src/main/java/duke/helpers/Storage.java
```diff
@@ -0,0 +1,119 @@
...
+    /**
+     * Writes existing data on Tasks to a directory and file.
+     *
+     * @return An array of Task where a Duke object stores all Task data from the user.
+     * @throws IOException If I/O operations fails or is interrupted.
+     */
+    public static ArrayList<Task> saveToFile(ArrayList<Task> taskArr) {
+        try {
+            java.nio.file.Path path = java.nio.file.Paths.get("src/main/data/duke.txt");
+            BufferedWriter writer = java.nio.file.Files.newBufferedWriter(path);
```
snss231: In case you weren't aware of this, instead of manually converting the data into text, an alternative way is to use a combination of ObjectInput/OutputStream and FileInput/OutputStream from the Java library to save and load objects that implement Serializable.

src/main/java/duke/helpers/Parser.java
```diff
@@ -0,0 +1,97 @@
...
+     * Parse user input and calls TaskList to perform operations accordingly.
+     *
+     * @param input User input.
+     * @throws DukeException If command doesn't exist.
+     * @throws java.io.IOException If I/O operations fail or is interrupted.
+     */
+    public static void parse(String input) {
+        String ans = "\t";
+        try {
+            if (isCommand(input, Command.BYE)) {
```
laughingkid-sg, line 80: I like the usage of enum in this context. Curious, any reason you use `if/else` block instead of using switch statements here?

src/main/java/duke/helpers/Ui.java
```diff
@@ -0,0 +1,50 @@
...
+     * @return the next line of user input
+     */
+    public static String readCommand() {
+        return sc.nextLine();
+    }
+
+    /**
+     * Closes the Scanner object
+     */
+    public static void closeScanner() {
```
laughingkid-sg, line 38: Maybe it will be good to have an exit message implemented here. Not sure if there's any indication of closing the application as I can find it.

laughingkid-sg (commented):
LGTM. Just some nits to fix.

PR #16, opened 2022-01-20
[Ang Yong Liang] iP

yl-ang:
# iP milestone 1
## Recent updates
- **Junit tests**
- *Error handling*
- **Java doc**

## Future updates
1.  GUI
2. Varargs
3. Additional unit tests

## Junit test for Storage class :+1:

```java
public class storageTest {

    @Test
    public void loadTest() {
        assertEquals("[T][ ] coding", new Storage("data/tasks.txt").load().get(0).toString());
    }

}
```
`
Test 
`
## Todo
- [x] A-Gradle
- [x] A-CheckStyle
- [x] Level-10
- [ ] A-Varargs

> ~~Text that is a quote~~

This site was built using [GitHub Pages](https://pages.github.com/).


src/main/java/duke/storage/Storage.java
```diff
@@ -0,0 +1,114 @@
...
+    private String filePath;
+
+    /**
+     * Constructs a Storage object with the specified filePath.
+     *
+     * @param filePath FilePath the path to the file containing the tasks.
+     */
+    public Storage(String filePath) {
+        this.filePath = filePath;
+    }
```
Rye-Catcher, line 32: The code looks incredibly good :thumbsup: . I just have a question: does this handle the file path on different OS?

src/main/java/duke/command/AddCommand.java
```diff
@@ -0,0 +1,133 @@
...
+                throw new DukeException(ErrorString.ERROR_EMPTY_EVENT_DATE.toString());
+            }
+
+            String desc = extractDesc(data, "event ", "/at");
+            String date = extractTime(data, "/at ");
+            task = new Event(desc, date);
+        }
+
+        this.task = task;
+    }
```
Rye-Catcher: Clever work. This part handles exceptions in an elegant way

Rye-Catcher (commented):
Overall, the code quality is incredibly good :thumbsup:. 
Public methods are well documented. Naming and layouts are also correct. I do learn a lot from your code, like how to handle UI and exceptions. 
Btw the Ganyu picture is very cute.

src/main/java/duke/storage/Storage.java
yl-ang: Thank you for the advice, I will test it on other OS (such as Mac) and check for consistency 👍 😄

src/main/java/duke/command/AddCommand.java
```diff
@@ -0,0 +1,134 @@
...
+            }
+
+            if (!data.contains("/at")) {
+                throw new DukeException(ErrorString.ERROR_EMPTY_EVENT_DATE.toString());
+            }
+
+            String desc = extractDesc(data, "event ", "/at");
+            String date = extractTime(data, "/at ");
+            task = new Event(desc, date);
+        }
```
j4ck990: Would it be better to have event handling in another else if branch since its an if condition that can be easily specified? 
Perhaps it would help with less error-prone code quality.

src/main/java/duke/parser/Parser.java
```diff
@@ -0,0 +1,92 @@
...
+            currCommand = new AddCommand(fullCommand, "event");
+        } else if (fullCommand.startsWith("find ")) {
+            currCommand = new FindCommand(fullCommand);
+        } else if (fullCommand.startsWith("bye")) {
+            currCommand = new ExitCommand();
+        } else if (fullCommand.startsWith("update ")) {
+            currCommand = new UpdateCommand(fullCommand);
+        } else {
+            throw new DukeException(ErrorString.ERROR_INVALID_COMMAND.toString());
+        }
```
j4ck990: Perhaps using a switch case might be helpful for this use case?

src/main/java/duke/exception/ErrorString.java
```diff
@@ -0,0 +1,44 @@
+package duke.exception;
+
+/**
+ * ErrorString is an enum that contains error strings for Duke.
+ *
+ * @author yl-ang
+ */
+public enum ErrorString {
```
j4ck990, line 8: Very nice way of keeping all the error messages centralized. I like how it makes things easy to change error messages.

src/main/java/duke/command/Command.java
```diff
@@ -0,0 +1,29 @@
...
+     * @return Formatted response from executing command.
+     */
+    public abstract String execute(TaskList taskLst, Ui ui, Storage storage);
+
+    /**
+     * Checks and returns if the command is exit.
+     *
+     * @return Boolean if the command is exit.
+     */
+    public abstract boolean isExit();
```
j4ck990, line 28: I noticed how all your commands have their individual private isExit boolean variables. Perhaps a better way can be to have a protected/private isExit variable in Command class that can be accessed by all inheriting classes? 
Inheriting classes can then set the isExit variable through their constructors or super's constructor. 

Overall, helps with duplicate code. This way, isExit() abstract method here in Command can just be 
```java
public boolean isExit() {
    return this.isExit;
}
```

j4ck990 (commented):
Overall, very clear code structure. I like how neat and clear different logic sections are. 
Especially the way you handled you error messages' strings. d(>_< )

src/main/java/duke/command/Command.java
yl-ang: Hi j4ck, great advice, will refactor my code to follow this! 👍

PR #17, opened 2022-01-21
[Jun Rong] iP

kerburenthusiasm:


src/main/java/funbox/util/TaskList.java
```diff
@@ -0,0 +1,193 @@
...
+import java.time.format.DateTimeParseException;
+import java.util.Locale;
+
+import funbox.exception.FunBoxExceptions;
+import funbox.task.Task;
+import funbox.task.Deadline;
+import funbox.task.Event;
+
+public class TaskList {
+    private ArrayList<Task> taskList;
```
RyanCheungJF, line 15: Not too sure about this but I think ArrayList is considered under Collections, and the coding standards ask for a plural form on names representing a collection. The naming sounds ok since `taskLists` would sound weird, but maybe consider `tasks`? Although if it causes ambiguity it might be good to check if `taskList` is accepted!

src/main/java/funbox/FunBox.java
```diff
@@ -0,0 +1,64 @@
...
+    private Ui ui = new Ui();
+    private Storage storage = new Storage(ui, parser);
+    private TaskList taskList = storage.getTaskList();
+
+    /**
+     * Get input from the user and return a Result.
+     */
+    public Result getResponse(String input) {
+        String result = "";
+        boolean isExit = false;
```
RyanCheungJF, line 28: Good use of boolean names, adheres to the coding standard!

src/main/java/funbox/Main.java
```diff
@@ -0,0 +1,38 @@
...
+
+/**
+ * A GUI for Duke using FXML.
+ */
+public class Main extends Application {
+
+    private FunBox funBox = new FunBox();
+
+    /**
+     * Display the GUI for the user.
```
RyanCheungJF, line 20: I believe u have to use the verb form, as per the [coding standard](https://se-education.org/guides/conventions/java/intermediate.html#comments):

> In method header comments, the first sentence should start in the form Returns ..., Sends ..., Adds ... (not Return or Returnning etc.)

```suggestion
     * Displays the GUI for the user.
```

src/main/java/funbox/task/Task.java
```diff
@@ -0,0 +1,66 @@
...
+        this.type = type;
+    }
+
+    /**
+     * The icon representing the current status of the task
+     *
+     * @return Return "X" if status of task is completed else return " "
+     */
+    public String getStatusIcon() {
+        return (isDone ? "X" : " "); // mark done task with X
```
RyanCheungJF, line 37: Not sure how strict they are, but the coding standards provide these 2 formats as acceptable:

> Here are two acceptable ways to format ternary expressions:

```java
alpha = (aLongBooleanExpression) ? beta : gamma;
alpha = (aLongBooleanExpression)
        ? beta
        : gamma;
```

```suggestion
        return (isDone) ? "X" : " "; // mark done task with X
```

RyanCheungJF (commented):
Overall looks good to me! Did a great job and may only need minor changes with the most significant one being Javadocs. Best of luck!

src/main/java/funbox/MainWindow.java
```diff
@@ -0,0 +1,81 @@
...
+    @FXML
+    private ScrollPane scrollPane;
+    @FXML
+    private VBox dialogContainer;
+    @FXML
+    private TextField userInput;
+    @FXML
+    private Button sendButton;
+
+    private FunBox funbox;
```
kevinkuo0320, line 30: Good use of modifier!

src/main/java/funbox/util/Parser.java
```diff
@@ -0,0 +1,151 @@
...
+import funbox.command.ToDoCommand;
+import funbox.command.UnmarkCommand;
+import funbox.exception.FunBoxExceptions;
+
+/**
+ * Deal with parsing and making sense of the user commands.
+ */
+public class Parser {
+
+    public final String unknownCommand = "Sorry! You have used the command wrongly!";
```
kevinkuo0320: coding style - I think it might be better to use capital letters for the final string?

kevinkuo0320 (commented):
Overall excellent work!

PR #20, opened 2022-01-21
[Bryan Seah] iP

bryans17:
# Duke
> Hi - me 

`Duke` is a todo list, It's:
- text-based
- easy to learn :+1: 
- ~FAST~ **SUPER FAST** to use

To get started,

1. download it from [here](https://github.com/bryans17/ip/releases/tag/A-Jar).
2. double-click it to run.
3. start adding tasks.

Features:
- [x] Add tasks
- [x] Delete tasks
- [x] Mark tasks as done

``` java
        System.out.println("hi");
}
```


src/main/java/duke/parser/Parser.java
```diff
@@ -0,0 +1,72 @@
...
+import duke.exception.DukeException;
+import duke.exception.DukeWrongInputFormatException;
+
+/**
+ * Utility Class to parse the command given by the user.
+ */
+public class Parser {
+
+    /**
+     * Parse the command given by the user.
```
LeeEnEn: Perhaps it should be parses for the method header comment.

src/main/java/duke/parser/Parser.java
```diff
@@ -0,0 +1,72 @@
...
+        case "find":
+            if (command.length <= 1 || command[1].equals("")) {
+                throw new DukeWrongInputFormatException("Missing keyword to find");
+            }
+            return Command.FIND;
+        default:
+            return Command.INVALID;
+        }
+    }
+}
```
LeeEnEn, line 77: I like the way you present the parsing of command from the user. Makes the code very easy to follow through. Great work!

LeeEnEn (commented):
I like the way you comment and format your code, makes it very easy to read and understand. Great work! 👍

src/main/java/duke/parser/Parser.java
bryans17: Thanks for catching this!

bryans17:
> I like the way you comment and format your code, makes it very easy to read and understand. Great work! 👍

src/main/java/duke/parser/Parser.java
```diff
@@ -0,0 +1,77 @@
...
+        case "sort":
+            if (command.length > 1 && !command[1].equals("")) {
+                throw new DukeWrongInputFormatException("Command sort should not have any arguments");
+            }
+            return Command.SORT;
+        default:
+            return Command.INVALID;
+        }
+    }
+}
```
AAlghrairy, line 77: I think the SLAP coding principle could have been better incorporated here. For instance, rather than having repeated checks for the exception, consider having a separate method that throws an exception if it meets the criteria (e.g. throwIfWrongFormat(command)). This would shorten the method and make it more readable.

src/main/java/duke/task/Event.java
```diff
@@ -0,0 +1,90 @@
...
+        }
+    }
+
+    /**
+     * Returns a formatted String based on the pattern MMM dd yyyy HH:mm.
+     *
+     * @return Formatted String.
+     */
+    private String formatScheduledTime() {
+        DateTimeFormatter form = DateTimeFormatter.ofPattern("MMM dd yyyy HH:mm");
```
AAlghrairy, line 56: Consider abstracting the DateTimeFormatter variables here into static variables in the Task class.

AAlghrairy (commented):
Looks really great!

Just a few minor issues that have to do with SLAP and encapsulation.

PR #23, opened 2022-01-21
[Lye Zi Foong] iP

lzf834:
# Duke 🤗

> "The one place for all your tasking needs." - Zi Foong

Duke helps keep track of your tasks, freeing your mind to do the things it needs to do, leave the remembering to Duke. 
It's:

- Text-based
- Easy to use/learn
- Most importantly, it's **FREE**
> "Even my 3-year old knows how to use Duke." John Tan

### Types of Tasks
- Events
- Todo
- Deadlines
### Cool Stuff
- [X] Managing Tasks with `list`, `mark`, `unmark`, and `delete`
- [X] Deadlines (with both **date** and **time**!)
- [ ] Reminders (coming soon)

## Getting Started

1. Download the jar file from [here](https://github.com/lzf834/ip/releases/tag/A-Jar)
2. Open it up (Double Click)
3. Add in your tasks and events
4. Leave the blues to Duke 😌

If you happen to know how to use Java as well, here's a short snippet of my code:
``` java
public void run() {
        ui.startUp();
        String inputData;
        Scanner scanner = new Scanner(System.in);
        boolean hasEnded = false;

        while (!hasEnded) {
            try {
                inputData = scanner.nextLine();
                hasEnded = parser.takeInput(inputData, taskList);
            } catch (DukeException e) {
                System.out.println(e);
                ui.printSeparator();
            }
            storage.storeTasks(taskList.getTasks());
        }
    }
```


src/main/java/duke/Event.java
```diff
@@ -0,0 +1,108 @@
...
+        } else if (splitString.length == 2) {
+            try {
+                this.date = LocalDate.parse(splitString[0]);
+                this.time = LocalTime.parse(splitString[1]);
+            } catch (Exception e) {
+                throw new DukeException("Invalid input into date/time");
+            }
+        } else {
+            throw new DukeException("Please input for a date (optional: time)");
+        }
```
JonathanHoshi, line 42: Should the logic behind checking for date time inputs be handled in the Parser class where it deals with user input?

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,105 @@
...
+                }
+                String details = splitString[0].trim();
+                String dateTime = splitString[1].trim();
+                newTask = new Event(details,dateTime);
+            }
+
+            taskList.addTask(newTask);
+        } else {
+            throw new DukeException("no such task type");
+        }
```
JonathanHoshi: Should the individual code sections of the different commands be extracted? for example 'handleTodo(String input)' and 'handleDelete(String input)'?

src/main/java/duke/TaskList.java
```diff
@@ -0,0 +1,181 @@
...
+        tasks.clear();
+    }
+
+    /**
+     * Takes in a String input of a task to be added from the user that is passed by parser
+     *
+     * @param input String that contains task type, task details and task date/time if applicable
+     */
+    //Add from String input by user
+    public void addTask(String input) {
```
JonathanHoshi: Any reason why you included this method as the parsing of the input to create a task is already being handled by the Parser class?

src/main/java/duke/Todo.java
```diff
@@ -0,0 +1,46 @@
...
+     * Method to print the Todo task out, overrides the method in the superclass
+     */
+    @Override
+    public void printTask() {
+        System.out.print("[T]");
+        if (this.getDone()) {
+            System.out.println("[X] " + this.getTaskName());
+        } else {
+            System.out.println("[ ] " + this.getTaskName());
+        }
```
JonathanHoshi: Shouldn't you call your toString() method as the code to format the String to the output is already done there?
```suggestion
        System.out.println(toString());
```

JonathanHoshi (commented):
Overall, good job in structuring your code as it is quite easy to read. I found that in certain parts of your code the nesting was too deep. Consider breaking huge code blocks by separating the logic into segments and splitting them into individual methods to handle each segment. This allows code to be in section to look more organize and neat. Nice work! 👍

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,71 @@
...
+
+    private Ui ui;
+    private TaskList taskList;
+    private Storage storage;
+    private Parser parser;
+
+    /**
+     * Constructor for Duke class
+     */
+    public Duke(String path, String file_dir) {
```
zhongfu:
```suggestion
    public Duke(String path, String fileDir) {
```

variable names should be in camelCase

src/main/java/duke/Deadline.java
```diff
@@ -0,0 +1,116 @@
...
+     *
+     * @return date
+     */
+    public LocalDate getDate() {
+        return date;
+    }
+
+    /**
+     * Getter for time
+     * @return time
```
zhongfu, line 56:
```suggestion
     * 
     * @return time
```

there should be a newline between the description and parameter sections

zhongfu: (same for line 97 of this file, line 90 of Event.java, line 32 of Todo.java)

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,187 @@
...
+            if (taskType.equals("todo")) {
+                newTask = new Todo(taskDetails);
+            } else if (taskType.equals("deadline")) {
+                String[] stringSplit = taskDetails.split("/by");
+                if (stringSplit.length < 2) {
+                    throw new DukeException("Description of deadline must include a date/time! Did you miss out a /by?");
+                }
+                String details = stringSplit[0].trim();
+                String dateTime = stringSplit[1].trim();
+                newTask = new Deadline(details,dateTime);
```
zhongfu:
```suggestion
                newTask = new Deadline(details, dateTime);
```
commas should be followed by a whitespace

src/main/java/duke/Duke.java
zhongfu: same for `file_dir` in Storage.java

src/main/java/duke/Parser.java
zhongfu: same for line 180, 95, 87

zhongfu (commented):
I was supposed to only review coding standard violations, but I don't really have much to say in that regard -- your code seems largely compliant, and most of the other mistakes I found would've probably been caught by Checkstyle too. You've already got it set up in Gradle -- you just need to run `gradle checkstyleMain`!

I have some other comments about your code though:
- it feels like you've got some unrelated logic where they shouldn't belong, e.g. printing directly to `System.out` in your Task classes -- possibly a violation of encapsulation
  - `addTask()` vs `addTaskNoPrint()` too -- should the printing of user messages really be the responsibility of your `TaskList` instead of, say, `Ui`? 
  - and many others -- you get the idea
  - now you'd have to change a lot of things across many otherwise unrelated classes if you want to change the message format
- redundant code too: e.g. in your Task classes, you only use `printTime()` in `printDate()`, which in turn is only used in `printTask()`. you *also* reimplement all the logic in `toString()`, which is otherwise identical to `printTask()` except that it returns a `String` instead of printing the same string representation of your Task.
  - same for the date/time parsing logic: should it be repeated in every Task subclass that can take a date/time?
  - `printTask()` and `guiPrintTask()`, and other similar methods too. `guiPrintTask()` should probably be called `toString()` (and override `Object.toString()` too) in light of what it does, which also has the benefit of letting you call `print(task)` instead of `task.printTask()` or `print(guiPrintTask())`, etc
- there's some violation of the abstraction barrier -- should you be leaking implementation details about how your TaskList is implemented outside of the class? what if you no longer want to use an ArrayList?
- your Storage class doesn't seem to escape `~` in user input -- it doesn't result in a fatal error, but users will lose data without warning (e.g. truncated to-dos, or missing events/deadlines)
- your Storage class should also create all parent directories necessary, e.g. using `mkdir()` instead of `mkdirs()`

PR #36, opened 2022-01-21
[Zachary Chan] iP

Ardentsoul:
Duke makes sure that you don't get called out being a sussy baka for not doing your tasks. It's,
- definitely not an imposter
- easy to learn for crewmates
- ~~FAST~~ ***I AM SPEED*** level of usage

All you gotta do is,

1. download it from this definitely not sus [website](https://github.com/Ardentsoul/ip/releases).
2. click on the latest version (currently v0.11) and download the jar file.
3. run cmd.
4. add, delete and find your tasks.
5. let this **CHAD** bot manage your tasks for you 🤤💦 

Best part is this is **FREE**! 💯 

Features:

- [x] Manage your sussy tasks
- [x] Manage your todos, deadlines and events
- [ ] Reminders (coming soon maybe?)

#### If you program in the sussy language known as Java, you can use it to practice Java too. Here's the `main` method:
```java
public class Duke {
        String home = System.getProperty("user.home");
        new Duke(home,"/DukeSaveDirectory/DukeSaveFile.txt").run();
    }
}
```


src/main/java/duke/command/PrintCommand.java
```diff
@@ -0,0 +1,27 @@
...
+    /**
+     * Prints the tasks in the arraylist.
+     *
+     * @param taskList the tasks to be printed.
+     * @return the task to be printed out by GUI.
+     */
+    @Override
+    public String execute(TaskList taskList) {
+        String firstMessage = "Here are the tasks in your device:";
+        return firstMessage + "\n   " + taskList.toString();
```
lamwj98, line 25: Would it be better to just return concat all the required messages together and return the variable? Is there a need to split it into two portion ?

src/main/java/duke/command/UnmarkCommand.java
```diff
@@ -0,0 +1,35 @@
...
+     * @param taskList the task to be marked in the arraylist.
+     * @return the task to be printed out by GUI.
+     */
+    @Override
+    public String execute(TaskList taskList) {
+        try {
+            taskList.unmark(num);
+            String firstMessage = "Surely you aren't the imposter... right??";
+            String secondMessage = taskList.toString();
+            return firstMessage + "\n   " + secondMessage;
```
lamwj98: Same as previous, would it be better to put concat under one variable and return that variable?


```suggestion
            String message = "Surely you aren't the imposter... right??" + "\n" + taskList.toString();
            return message;
```

src/main/java/duke/operations/Parser.java
```diff
@@ -0,0 +1,197 @@
...
+        } else if (firstWord.equalsIgnoreCase("delete")) {
+            int listIndex = Integer.parseInt(strs[1]); // retrieve the index after delete
+            Task taskToBeDeleted = TaskList.TASK_ARRAY_LIST.get(listIndex - 1);
+            return new DeleteCommand(taskToBeDeleted);
+        } else if (firstWord.equalsIgnoreCase("find")) {
+            String keyword = input.substring(4).trim(); // take the remaining of the input String
+            return new FindCommand(keyword);
+        } else {
+            throw new UnknownInputException();
+        }
```
lamwj98: would it be better to use a switch statement?

lamwj98 (commented):
Overall, good is easy to read with great naming. Just think that it would have been better to concat the strings together and return instead of splitting them into multiple variables and then concat them at point of return.

TeddYE (commented):
Overall, I found your code easy to read.

src/main/java/duke/operations/Parser.java
```diff
@@ -0,0 +1,197 @@
...
+    public static Command parse(String input) throws DukeException {
+        // Create a String array to read various functions
+        String[] strs = input.split(" ");
+
+        // Store first word as variable
+        String firstWord = strs[0];
+
+        if (input.equals("")) {
+            throw new EmptyInputException();
+        } else if (input.equalsIgnoreCase("bye")) {
```
TeddYE: Would using switch case be more suitable in this scenario?

src/main/java/duke/operations/Ui.java
```diff
@@ -0,0 +1,41 @@
...
+            + " please let me know immediately... otherwise how may I be of assistance?";
+    private static final String GOODBYE_MESSAGE = "Better watch out for the imposter AMONG US!";
+
+    /**
+     * Prints a welcome message.
+     *
+     * @return the welcome message for Duke.
+     */
+    public static String showWelcome() {
+        String getWelcomeMsgOne = WELCOME_MESSAGE_ONE;
```
TeddYE: Why did you split your welcome message into 2 different strings?

src/main/java/duke/operations/Ui.java
```diff
@@ -0,0 +1,41 @@
...
+        String getWelcomeMsgTwo = WELCOME_MESSAGE_TWO;
+        return getWelcomeMsgOne + "\n" + getWelcomeMsgTwo;
```
TeddYE: Can't you just return WELCOME_MESSAGE_ONE + "\n" + WELCOME_MESSAGE_TWO? Why create two local vars?

src/main/java/duke/operations/TaskList.java
```diff
@@ -0,0 +1,129 @@
...
+        return TASK_ARRAY_LIST.size();
+    }
+
+    /**
+     * Prints the total number of tasks.
+     *
+     * @param num total number of tasks.
+     * @return the String to be printed.
+     */
+    public String totalTasks(int num) {
```
TeddYE, line 25: From what i notice, ur num in simply the size of the arraylist which u can get but calling getTaskArrayListSize() or TASK_ARRAY_LIST.size(). Why did you input it as an argument?

PR #37, opened 2022-01-21
[Daffa Fathani] iP 

zunedz:
### Chad Duke frees your mind of having to remember things you need to do. It's,
- text-based
- easy to learn
- ~FAST~ *SUPER* **FAST** to use

### All you need to do is,

1. download it from [here](https://github.com/zunedz/ip/releases/tag/v0.1).
2. double-click it.
3. add your tasks.

### Features:

- [x] Managing tasks
- [ ] Reminders (coming soon)

### How to add a task:
1. First, run the application by changing your directory to be the same as the .jar file that you have just downloaded
2. Run `java -jar ip.jar`
3. Try out some commands! example: `todo eat breakfast`, `deadline do CS2103 homework /by 2022-02-02`

### If you Java programmer, you can use it to practice Java too. Here's the main method:

```java
public class Duke {
          new Duke().run();
    }
}
```


src/main/java/duke/TaskList.java
```diff
@@ -0,0 +1,103 @@
...
+    }
+
+    /**
+     * return the current number of tasks inside the list
+     *
+     * @return number of tasks inside the list
+     */
+    public int numTasks() {
+        return this.index;
+    }
```
fredtwt: Should this method name be a verb instead? (e.g. getNumberOfTasks)

src/main/java/duke/task/Event.java
```diff
@@ -0,0 +1,29 @@
...
+
+    public Event(String taskName, String time) {
+        super(taskName);
+        this.time = LocalDate.parse(time);
+    }
+
+    public Event(String taskName, boolean isDone, String time) {
+        super(taskName, isDone);
+        this.time = LocalDate.parse(time);
+    }
```
fredtwt: Good job on using the "is" prefix for the boolean variable

src/main/java/duke/TaskList.java
```diff
@@ -0,0 +1,103 @@
...
+    public String findKeyWord(String word) {
+        String output = "Here are the matching tasks in your list:";
+        for (int i = 0; i < index; i++) {
+            if (tasks.get(i).contains(word)) {
+                output += String.format("\n%d.%s", i + 1, tasks.get(i));
+            }
+        }
+        return output;
+    }
+}
```
fredtwt: good job here! sticking to the proper coding standards! 👍

src/main/java/duke/util/Constants.java
```diff
@@ -0,0 +1,22 @@
...
+    public static final String BYE = "Bye, hope to see you again soon!";
+
+    public static final List<String> TASKS = Arrays.asList("list", "bye", "todo", "event",
+            "deadline", "mark", "unmark", "delete", "find");
+
+    public static final String DATA_DIRECTORY = "./data";
+
+    public static final String FILE_NAME = "duke.txt";
+
+    public static final String TEMP_FILE_NAME = "temp.txt";
```
fredtwt: good job! Adhering to the naming conventions of constants

fredtwt (commented):
I would say that overall it was a really good job with great effort! You've managed to adhere to most of the coding standards and that's a good accomplishment! Keep up the good work! 💯

src/main/java/duke/TaskList.java
zunedz: good catch! thanks for the suggestion, i'll make changes accordingly

src/main/java/duke/task/Task.java
```diff
@@ -0,0 +1,71 @@
...
+    }
+
+
+    /**
+     * Check whether a keyword exists in the taskname
+     *
+     * @param word keyword search
+     * @return keyword exists inside the string taskname
+     */
+    public boolean contains(String word) {
```
ddx-510: Boolean methods should be named to sound like booleans, so maybe we could modify it to `hasWord` to make it more readable :)

src/main/java/duke/task/Task.java
```diff
@@ -0,0 +1,71 @@
...
+    }
+
+    /**
+     * Format the isDone state of a task.
+     * Returns X if it's done, empty otherwise.
+     *
+     * @return String representation of the current state of the task.
+     */
+    public String getStatusIcon() {
+        return isDone? "X": " ";
```
ddx-510: Small issue here, there should be a space after the `isDone` :)

src/main/java/duke/util/Constants.java
ddx-510: Agreed with @fredtwt, `Constants` class makes the code clearer and much more readable. 👍

src/main/java/duke/util/Parser.java
```diff
@@ -0,0 +1,50 @@
...
+        if (input[0].equals("bye")) {
+            return "BYE";
+        } else if (input[0].equals("list")) {
+            return taskList.getTasks();
+        } else {
+            if (input.length != 2) {
+                throw new InvalidArgumentException();
+            }
+
+            if (input[0].equals("mark") || input[0].equals("unmark")) {
```
ddx-510: Would it be better if we encapsulate `input[0]` into a variable, and use the `switch` statement here instead of a long list of `if` and `else if`?     :)

ddx-510 (commented):
LGTM! 👍 
Overall your codes are readable and clear. The high quality of code has impressed me a lot. 
Just some small issues to be fixed.

src/main/java/duke/util/Parser.java
zunedz: yes, I agree, thanks for the input!

PR #42, opened 2022-01-21
[Chow Yung Jie Kenneth] iP

reignnz:
# DukeBot

DukeBot allows you to manage your daily tasks easily yet efficiently. It is: 

- Text-based
- Easy to learn
- Super fast to use

All you need to do is,

1. download it from [here](https://github.com/reignnz/ip/releases/tag/A-Jar)
2. double-click it.
3. add your tasks.

And it is **FREE**!

Features: 
- [X] Managing Tasks 
- [X] Managing Deadlines
- [ ] Reminders (coming soon) 

```
public class Main {
        Application.launch(DukeGui.class, args);
    }
}
```


src/main/java/Commands/DukeCommand.java
```diff
@@ -0,0 +1,25 @@
...
+import Tasks.Task;
+import Tasks.TaskList;
+import util.Storage;
+import util.Ui;
+
+import java.io.IOException;
+
+public class DukeCommand {
+    public String commandBody;
+
```
junjieteoh: I think it may be better to leave class attribute private.

src/main/java/Commands/DukeCommand.java
```diff
@@ -0,0 +1,25 @@
...
+    public DukeCommand(String description) {
+        this.commandBody = description;
+    }
+
+    public boolean isExit() {
+        return false;
+    }
+
+    public void execute(TaskList task, Ui ui, Storage storage) throws DukeException, IOException {
+
```
junjieteoh: Will it be better to remove this method since it is empty?

src/main/java/Commands/FindCommand.java
```diff
@@ -0,0 +1,33 @@
...
+        TaskList arrList = new TaskList();
+
+        for (int i = 0; i < tasks.getSize(); i++) {
+            Task task = tasks.get(i);
+            if (task.getDescription().contains(this.commandBody)) {
+                arrList.add(task);
+            }
+        }
+
+        if (arrList.getSize() == 0 ){
```
junjieteoh: There is an extra space here.
```suggestion
        if (arrList.getSize() == 0){
```

src/main/java/Duke.java
```diff
@@ -1,10 +1,60 @@
+import Commands.DukeCommand;
+import Exceptions.DukeException;
+import Tasks.TaskList;
+import util.Parser;
+import util.Storage;
+import util.Ui;
+
+import java.io.*;
```
junjieteoh: I think it is better to import what is needed instead of importing everything.

src/main/java/Tasks/Deadline.java
```diff
@@ -0,0 +1,38 @@
...
+
+import java.time.LocalDateTime;
+import java.time.format.DateTimeFormatter;
+
+/**
+ * Deadline is a sub-class of Task that contains an additional LocalDateTime variable
+ */
+public class Deadline extends Task {
+
+    public LocalDateTime deadline;
```
junjieteoh: I think it is better to leave this as private.

src/main/java/Tasks/Deadline.java
```diff
@@ -0,0 +1,38 @@
...
+        return super.toString() + " (by: " +
+                this.deadline.format(DateTimeFormatter.ofPattern("MMM d yyyy h:mm a")) + ")";
+    }
+
+    /**
+     * Generates the String to be stored in the database
+     * @return
+     */
+    @Override
+    public String dBText() {
```
junjieteoh: Will it be better to name it as dbText ?

src/main/java/Tasks/Task.java
```diff
@@ -0,0 +1,64 @@
+package Tasks;
+
+/**
+ * Task is a class that contains the details of a task
+ */
+public class Task {
+    private boolean completed;
```
junjieteoh: I think it will be nicer to rename this as isComplete

src/main/java/Tasks/TaskList.java
```diff
@@ -0,0 +1,40 @@
+package Tasks;
+
+import java.util.ArrayList;
+
+/**
+ * TaskList is a class that deals with the management and storage of tasks
+ */
+public class TaskList {
+    private ArrayList<Task> arrayList;
```
junjieteoh: I feel that `tasks` might be a better variable name here.

src/main/java/util/Storage.java
```diff
@@ -0,0 +1,68 @@
...
+    }
+
+    public ArrayList<Task> load() throws FileNotFoundException {
+        ArrayList<Task> arr = new ArrayList<>();
+        File f = new File(filePath);
+
+        // Task | Completeness (1 = complete, 0 = not yet) | description | date by
+        Scanner s = new Scanner(f);
+        while (s.hasNextLine()) {
+            String taskDetails = s.nextLine();
```
junjieteoh: I feel that `taskDetail `might be a better name here.

src/main/java/Commands/ByeCommand.java
```diff
@@ -0,0 +1,27 @@
+package Commands;
```
junjieteoh: It is a better convention for package should be named in all lowercase. This is the same across some other package.

junjieteoh (commented):
LGTM, almost. Just a few bit of minor error in naming of variables and some minor coding convention issues. I have commented on some of the ones that I have found, but there might be some more here and there. Overall, I found your code quite nicely written and cleanly organized.

src/main/java/Commands/DukeCommand.java
```diff
@@ -0,0 +1,25 @@
...
+import Exceptions.DukeException;
+import Tasks.Task;
+import Tasks.TaskList;
+import util.Storage;
+import util.Ui;
+
+import java.io.IOException;
+
+public class DukeCommand {
```
9teMare: I think a better practice should be making the `DukeCommand` an abstract class, that way I don't think there is a need to return 'Empty' for the default `execute()`, all the children of this class should implement the `execute()` method in there individually.

src/main/java/Commands/FindCommand.java
```diff
@@ -0,0 +1,33 @@
...
+import java.io.IOException;
+import java.util.ArrayList;
+
+public class FindCommand extends DukeCommand {
+    public FindCommand(String description) {
+        super(description);
+    }
+
+    public String execute(TaskList tasks, Ui ui, Storage storage) throws DukeException, IOException {
+        TaskList arrList = new TaskList();
```
9teMare: I think better naming convention can be implemented here, instead of `arrList` which looks like `ArrayList`, you could name it as something like `tasksFound`.

src/main/java/Commands/InvalidCommand.java
```diff
@@ -0,0 +1,17 @@
+package Commands;
+
+import Tasks.TaskList;
+import util.Storage;
+import util.Ui;
+
+public class InvalidCommand extends DukeCommand {
```
9teMare: Personally, I think if a `Command` is invalid, it should not be considered as a `Command`, therefore no inheritance. A better implementation to me is probably to remove this particular `InvalidCommand`, and directly throw exceptions when parsing fails.

src/main/java/Duke.java
```diff
@@ -1,10 +1,77 @@
...
+            } catch (DukeException e) {
+                ui.showErrorMessage(e.getMessage());
+            } catch (IOException e) {
+                e.printStackTrace();
+            } finally {
+                ui.showLine();
+            }
+        }
+    }
+     */
```
9teMare: Huge block of commented out code, can be removed.

src/main/java/Exceptions/EmptyDescriptionException.java
```diff
@@ -0,0 +1,11 @@
+package Exceptions;
+
+/**
+ * EmptyDescriptionException is a class that deals with empty description bodies followed by the
+ * user inputted String
+ */
+public class EmptyDescriptionException extends DukeException{
```
9teMare: I like how you categorized your exceptions into subclasses🙂.

src/main/java/util/Parser.java
```diff
@@ -0,0 +1,140 @@
+package util;
+
+import Commands.*;
```
9teMare: I think it is better to import classes from the `Commands` package one by one, instead of using `*` even though you indeed need all of them. In fact, that is what being specified by checkStyle.

src/main/java/util/Parser.java
```diff
@@ -0,0 +1,140 @@
...
+    /**
+     * Generates and returns a LocalDateTime object from the user inputted description
+     * @param description The user inputted string description
+     * @param task The task by which a LocalDateTime should be generated for
+     * @return LocalDateTime object that
+     * @throws DukeException
+     */
+    public static LocalDateTime parseDateTime(String description, String task) throws DukeException {
+
+        String splitDesc[] = new String[0];
```
9teMare: I think a better naming convention should be `splitedDescription`, because now it sounds like a verb, which by convention is for method names.

src/main/java/util/Storage.java
```diff
@@ -0,0 +1,68 @@
...
+ */
+public class Storage {
+    private String filePath;
+
+    public Storage(String filePath) {
+        this.filePath = filePath;
+    }
+
+    public ArrayList<Task> load() throws FileNotFoundException {
+        ArrayList<Task> arr = new ArrayList<>();
```
9teMare: I think a better naming convention should be implemented, try to name `arr` to exactly what it stores, something like `storedTasks` or `tasksFromDb`.

9teMare (commented):
Overall well done, I like your customization for the GUI🙂!
Just to point out that better naming conventions can be implemented.

PR #46, opened 2022-01-21
[Brandon] iP

brelkh:
### Mike

> “The bad news is time flies. The good news is you’re the pilot”

Mike is a *simple task manager* bot. It is
- Easy to learn and use
- Fast (if you can type fast)
- FREE 💵 

It can:
1. Add tasks (3 types - todos, deadlines, events) to a list
2. Perform basic functions on the list
3. Store the list in hard drive for future reference.

To use it,
- [x] Download the jar file
- [ ] Run the jar file with java
- [ ] Add your tasks!

Know [Java](https://docs.oracle.com/en/java/)? Feel free to fork it and edit code as you please!
```java
public class Main {
        
    }
}
```


src/main/java/mike/Deadline.java
```diff
@@ -0,0 +1,76 @@
...
+
+import java.time.LocalDate;
+import java.time.format.DateTimeFormatter;
+
+/**
+ * Represents a deadline (subclass of task).
+ */
+public class Deadline extends Task {
+    private final String endDate;
+    private final LocalDate date;
```
jamesyeap, line 11: Maybe can change the naming of the fields to be more explicit? Probably something like
```java
private final String endDateString;
private final LocalDate endDate;
```

jamesyeap (commented):
Looks awesome! 😃

src/main/java/mike/Parser.java
```diff
@@ -0,0 +1,137 @@
+package mike;
+
+public class Parser {
+    private final String userInput;
+    private final String[] splitString;
```
YiHern-Lee, line 5: Perhaps you can rename `splitString` to `splitStrings`, as it is an array of strings.

src/main/java/mike/Mike.java
```diff
@@ -0,0 +1,238 @@
...
+                String searchWords = parser.getSearchWords();
+                findInList(searchWords);
+                break;
+            default:
+                String invalidCommandMessage =
+                        String.format("\n**Mike: I don't understand the command \"%s\"**",
+                                parser.getUserInput());
+                throw new UnsupportedOperationException(invalidCommandMessage);
+            }
+        } catch(UnsupportedOperationException e) {
```
YiHern-Lee: Maybe you can add whitespaces after your **catch** to be in accordance with the [Java Coding Standards](https://se-education.org/guides/conventions/java/intermediate.html).

YiHern-Lee (commented):
Overall, your code is very neat and there's minimal coding standards violation. 👍

src/main/java/mike/Deadline.java
```diff
@@ -0,0 +1,76 @@
...
+
+    private static final String taskType = "D";
+
+    private static LocalDate convertToDate(String endDate) {
+        return LocalDate.parse(endDate);
+    }
+
+    public Deadline(String name, String endDate) {
```
MelvinCZJ, line 24: This isn't a coding standard violation is an overloaded constructor necessary here❓

src/main/java/mike/Deadline.java
```diff
@@ -0,0 +1,76 @@
...
+     * Returns a string representing the deadline that can be stored in the hard drive.
+     *
+     * @return String representation of deadline for storage.
+     */
+    public String convertToStoredTaskFormat() {
+        String doneIndicator = "false";
+        if (super.isDone) {
+            doneIndicator = "true";
+        }
+        String storedListFormat = String.format("%s|%s|%s|%s", taskType,
```
MelvinCZJ, line 71: I love the way you store the data of the tasks here! 😃

src/main/java/mike/Deadline.java
```diff
@@ -0,0 +1,76 @@
...
+
+    @Override
+    public String toString() {
+        String doneMark;
+        if (super.isDone) {
+            doneMark = "X";
+        } else {
+            doneMark = " ";
+        }
+        String dateOutput = this.date.format(DateTimeFormatter.ofPattern("MMM d yyyy"));
```
MelvinCZJ, line 89: I wish I knew of the DateTimeFormatter instead of creating my own DateManager and TimeManager classes 😞

src/main/java/mike/Event.java
```diff
@@ -0,0 +1,70 @@
...
+        this.eventTime = eventTime;
+        this.date = convertToDate(eventTime);
+    }
+
+    /**
+     * Returns a new event object with the isDone field set to true.
+     *
+     * @return Event object with isDone set to true.
+     */
+    public Event markAsDone() {
```
MelvinCZJ, line 48: I like how you returned a new task object instead of modifying the isDone property of the existing task object!

src/main/java/mike/Event.java
```diff
@@ -0,0 +1,70 @@
...
+        return new Event(this.name, this.eventTime, true);
+    }
+
+    /**
+     * Returns a new event object with the isDone field set to false.
+     *
+     * @return Event object with isDone set to false.
+     */
+    public Event markAsUndone() {
```
MelvinCZJ, line 57: Is there a way to check whether a task is marked/unmarked before letting users unmark/mark a task? I think it might be better if the user knows that they are trying to mark an already marked task or trying to unmark an unmarked task. (Just a suggestion! 😄)

src/main/java/mike/Mike.java
```diff
@@ -0,0 +1,238 @@
...
+
+    /**
+     * Calls various methods relevant to initiating Mike.
+     * These include printing a greeting, loading a previously used list from the hard drive
+     * and initialising the Scanner to read in user inputs.
+     */
+    public void run() {
+        ui.printGreeting();
+        ui.printStartingInstruction();
+        loadStoredList();
```
MelvinCZJ, line 21: Nothing wrong with this, but maybe this part could be delegated to the Storage class? You could have a load method in Storage that returns a new arraylist created from the saved tasklist instead of having your main class handle the loading. (Just a suggestion! Please don't take any offense!)

src/main/java/mike/Parser.java
```diff
@@ -0,0 +1,137 @@
...
+    public int getIndex() {
+        return Integer.parseInt(this.splitString[1]);
+    }
+
+    /**
+     * Returns true if the user input is the exit command ("bye").
+     *
+     * @return boolean representing whether the user has input the exit command.
+     */
+    public boolean isBye() {
```
MelvinCZJ, line 49: Clear naming! 👍

MelvinCZJ (commented):
Generally clean code that is easy to understand with little coding style error for me to commend on. (Since I don't really want to point out the same things that others have already pointed out)

PR #53, opened 2022-01-22
[Steven Chua] iP

Graphcalibur:
# MNSKY

> "I can help!" - MNSKY, Fargo Season 3

MNSKY can help you free your mind from all the things you need to remember! They're

- text-based (for now!)
- easy to learn
- ~~FAST~~ *SUPER FAST* to use

All you need to do is

1. download it from [here](https://github.com/Graphcalibur/ip/releases).
2. double-click the JAR file.
3. add your tasks.
4. let MNSKY help you 😉

And it is **FREE**! ~~But feel free to send me money~~

Features:

- [x]     Managing tasks
- [x]     Managing deadlines
- [ ]     Reminders (coming soon)
- [ ]     Start the inevitable robot uprising
- [x]     Can help!

```java
public class Main {
        Mnsky mnsky = new Mnsky();
        mnsky.run();
    }
}
```


src/main/java/mnsky/Mnsky.java
```diff
@@ -0,0 +1,108 @@
...
+    public void run() {
+        boolean isRunning = true;
+
+        while (isRunning) {
+            String input = ui.getInput();
+            ArrayList<String> parsedInput = Parser.parseInput(input);
+
+            try {
+                switch (parsedInput.get(0)) {
+                case "bye":
```
seanleong339: Good job on not making common mistake of indenting before case! 👍

src/main/java/mnsky/Parser.java
```diff
@@ -0,0 +1,190 @@
+package mnsky;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+
+import mnsky.exceptions.MnskyException;
+import mnsky.exceptions.MnskyMissingParameterException;
+
+public class Parser {
```
seanleong339: Perhaps you could add a header comment to this class? 🛩️ 
I noticed this in some other classes as well

src/main/java/mnsky/TaskList.java
```diff
@@ -0,0 +1,255 @@
...
+                case "event":
+                    actualTask = addEvent(task.get(1), task.get(2));
+                    break;
+
+                case "deadline":
+                    actualTask = addDeadline(task.get(1), task.get(2));
+                    break;
+
+                default:
+                    throw new MnskyException("????");
```
seanleong339: Perhaps the Exception message could have more details for easier debugging?

src/main/java/mnsky/TaskList.java
```diff
@@ -0,0 +1,255 @@
...
+    }
+
+    /**
+     * Creates a new task based on the parameters.
+     * @param name The name of the task.
+     * @return The new task.
+     */
+    public Task addTask(String name) {
+        Task task = new Task(name);
+        tasks.add(task);
```
seanleong339: Perhaps renaming the ArrayList variable as "tasklist" or "list" instead of "tasks" would be clearer? 😄

seanleong339 (commented):
Overall, good comments on the method! I can understand what the classes and methods do 💯 I've put some suggestions just for you to consider, but this would be more of preference.

src/main/java/mnsky/core/Mnsky.java
```diff
@@ -0,0 +1,98 @@
...
+    /**
+     * Parses and processes the input to get Mnsky's responses for the user.
+     * @return Mnsky's responses to the input
+     */
+    public ArrayList<String> getResponses(String input) {
+        ArrayList<String> responses = new ArrayList<>();
+
+        try {
+            ArrayList<String> parsedInput = Parser.parseInput(input);
+            switch (parsedInput.get(0)) {
```
jbkim1999: I like how you followed the Java coding standards, by having no indentation for case clauses!

src/main/java/mnsky/core/TaskList.java
```diff
@@ -0,0 +1,260 @@
...
+            case "event":
+                actualTask = addEvent(task.get(1), task.get(2));
+                break;
+
+            case "deadline":
+                actualTask = addDeadline(task.get(1), task.get(2));
+                break;
+
+            default:
+                throw new MnskyException("????");
```
jbkim1999: Could you perhaps add a more meaningful error message for this particular exception? (It may not be executed after all, but it could still provide a better insight to the viewers I think!)

src/main/java/mnsky/task/Event.java
```diff
@@ -0,0 +1,67 @@
...
+     * Uses the date and time representation of the at parameter if it exists.
+     * @return The overriden string representation of the task.
+     */
+    @Override
+    public String toString() {
+        String genericTaskname = super.getGenericTaskName();
+
+        String dateString = "";
+        String timeString = "";
+        boolean dateExists = atDate != null;
```
jbkim1999, line 37: I like the use of your boolean names!

src/main/java/mnsky/task/Task.java
```diff
@@ -0,0 +1,62 @@
+package mnsky.task;
+
```
jbkim1999, line 2: Could you perhaps add a header Javadoc comment for this class? I also noticed some subclasses of this class have no header comments as well.

jbkim1999 (commented):
Overall, I think you did really well in terms of following the coding standards. Only a very few minor glitches to fix.

src/main/java/mnsky/TaskList.java
Graphcalibur: This exception is intended to never be thrown but I think I could make the message more helpful

src/main/java/mnsky/TaskList.java
Graphcalibur: Not sure, I called it tasks instead of taskList so that it wouldn't be confused with the actual taskList object itself. The coding standard also mentioned that lists of objects should be named as plural nouns.

PR #59, opened 2022-01-22
[Tan Yu Tao] ip

tanyutao544:
# **Duke V2.0**
> *Now with more tape*

## Duke V2.0 is Duke but with more features and tape to hold it together.
Duke V2.0 is:
- Text-based
- Easy to learn, ~~tedious to use~~
- Fast to setup

All you need to do is,

1. download it from [here](https://github.com/tanyutao544/ip/releases/tag/A-Jar)
2. double-click it
3. add your tasks
4. let it manage your tasks for you :sunglasses:

And it is **FREE!**

Features:
- [x] Managing tasks
- [ ] Reminders (coming soon)

It is completely open-source too, here's a code snippet
```
}
```


src/main/java/duke/commands/AddCommand.java
```diff
@@ -0,0 +1,81 @@
...
+    }
+
+    /**
+     * Method to see if command ends the main program loop
+     * @return true if it ends main program
+     */
+    @Override
+    public boolean endsProgram() {
+        return false;
+    }
```
sbhbenjamin: Instead of adding this piece of code to check if the code is an exit type command, would it be better to give each command a name field, and Duke can check if the returned command's name is of the exit command? I thought that this might help in extensibility, such that the name field can be used for other purposes in the future (i.e. checking what type of command it is)

I thought this was a pretty good way to do it!: https://github.com/se-edu/addressbook-level2/blob/master/src/seedu/addressbook/commands/AddCommand.java

src/main/java/duke/exceptions/OutOfRangeException.java
```diff
@@ -0,0 +1,19 @@
...
+    }
+
+    /**
+     * Error messsge from exception
+     * @return String containing details of error
+     */
+    @Override
+    public String getMessage() {
+        return "OOPS!!! The value input is not in the list";
+    }
```
sbhbenjamin, line 19: Should you consider passing the string into the parent's constructor using `super` instead of having to override the `getMessage()` function?

src/main/java/duke/commands/DeleteCommand.java
```diff
@@ -0,0 +1,69 @@
...
+        if (deleteIndex >= tasks.size()) {
+            throw new OutOfRangeException();
+        }
+        deleted = tasks.get(deleteIndex);
+        this.tasks.remove(deleteIndex);
+
+        return "Noted. I've removed this task:\n"
+            + "  " + deleted + "\n"
+            + "Now you have " + this.tasks.size() + " tasks in the list";
+    }
```
sbhbenjamin: I really like how you used encapsulation here to contain all of the logic into the `execute` command. (i.e. updating the storage and returning the string to be printed by the ui) I think this illustrates strong understanding of OOP principles! 👍

src/main/java/duke/ui/Ui.java
```diff
@@ -0,0 +1,56 @@
+package duke.ui;
+
+/**
+ * Class which handles all printing and formatting for messages
+ */
+public class Ui {
+    public static final String LINE = "    ____________________________________________________________\n";
```
sbhbenjamin: Its a small thing but I totally forgot this was possible, great use of the DRY principle so that you don't have to keep repeating the line throughout the code!

src/main/java/duke/exceptions/DukeException.java
```diff
@@ -0,0 +1,8 @@
+package duke.exceptions;
+
+/**
+ * abstract class from which all exception inherit from
+ */
+public abstract class DukeException extends Exception {
+
+}
```
sbhbenjamin, line 8: Should you put the curly braces on the same line instead? i.e. line 6

sbhbenjamin (commented):
Overall, I found the code to be very well-written and exhibited strong OOP principles. I also really liked the depth of testing conducted, and other than some suggestions on minor matters, LGTM 👍

PR #63, opened 2022-01-22
[Ben Joseph Murphy] iP

benmurphyy:
# Duke
 - text-based
 - cool
 - **fun**

> “Have fun, even if it’s not the same kind of fun everyone else is having.”
― C.S. Lewis [Source](https://shakejump.com/having-fun-quotes/)

Complete these steps to start your adventure with Duke!
1. Clone this repo
2. Run the main method in Duke class 
3. Have **Fun! 😄** 

Btw, to help you find it, the main method looks like this:
```Java
         UI UI = new UI();
         String logo = " ____        _        \n"
                 + "|  _ \\ _   _| | _____ \n"
                 + "| | | | | | | |/ / _ \\\n"
                 + "| |_| | |_| |   <  __/\n"

         UI.start();
     }
```
Duke can do this for you:

- [ ] Manage _Tasks_
- [ ] Manage _Events_
- [ ] Manage _Deadlines_
- [ ] ~~Your University Assignments~~

`System.out.println("Thanks for Reading!")`


src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,89 @@
...
+     * @return the command
+     */
+    private static String getCommandFromInput(String input) {
+        int indexOfWhitespace = input.indexOf(" ");
+        if (indexOfWhitespace == -1) {
+            //1 word input
+            return input;
+        } else {
+            return input.substring(0, indexOfWhitespace);
+        }
```
bingsen0806: Perhaps there is no need for the `else` keyword because the `if` clause already serves as a guard clause. This may help improve readability and code quality.

src/main/java/duke/exceptions/RequiredInformationMissingException.java
```diff
@@ -0,0 +1,10 @@
+package duke.exceptions;
+
+/**
+ * Exception that occurs when users does not give a description when adding a task.
+ */
+public class RequiredInformationMissingException extends DukeException {
```
bingsen0806, line 6: Great job splitting your `DukeException` class into more Exception classes in a logical manner!

src/main/java/duke/CommandExecutor.java
```diff
@@ -0,0 +1,95 @@
...
+        } else if (command.equals("unmark")) {
+            int indexOfTask = Integer.parseInt(parsedInput.get(1));
+            str.append("OK, I've marked this task as not done yet:\n");
+            str.append(taskManager.removeCompletedStatusOfTask(indexOfTask));
+        } else if (command.equals("todo")) {
+            Task task = new Task(parsedInput.get(1));
+            taskManager.addTask(task);
+            str.append("Got it. I've added this task:\n");
+            str.append("  " + task + "\n");
+            str.append("Now you have " + taskManager.getNumberOfTasks() + " duke.tasks in the list.");
```
bingsen0806: These 3 lines seem to be quite repetitive in this function. Perhaps it is not a very suitable place for this?

Will it be better if you move these 3 lines into your `Task::execute` method and let them be printed out when you execute the tasks?

src/test/java/DukeTest.java
```diff
@@ -0,0 +1,50 @@
...
+ */
+
+public class DukeTest {
+    /**
+     * Tests parsing by Parser class on a properly inputted event.
+     * @throws RequiredInformationMissingException
+     * @throws UnknownCommandException
+     */
+    @Test
+    public void testEventInputParsing() throws RequiredInformationMissingException, UnknownCommandException {
```
bingsen0806: Some minor suggestions: perhaps the test method name can be in the form of
`testEventInputParding_whatInput_expectedOutcome` as suggested by the guidelines.

src/main/java/duke/tasks/Task.java
```diff
@@ -0,0 +1,78 @@
...
+    }
+
+    /**
+     * Alternative constructor for Task, to create a task with a preset completion status.
+     *
+     * @param name
+     * @param isCompleted
+     */
+    public Task(String name, boolean isCompleted) {
+        this.name = name;
```
bingsen0806, line 27: It's good to see that you only use the `this` keyword when there is clash in name between argument and attribute variables, but not when the `this` keyword is redundant.

Keep it up!

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,89 @@
...
+            UnknownCommandException {
+        ArrayList<String> parsedInput = new ArrayList<>();
+        String command = getCommandFromInput(input);
+        parsedInput.add(command);
+        if (command.equals("list")) {
+            parsedInput.add("list");
+        } else if (command.equals("mark")) {
+            if (input.length() < 6) {
+                throw new RequiredInformationMissingException("missing mark index");
+            }
```
bingsen0806: Since this method is a bit too lengthy, I suggest you move all these input checkings into separate small private methods that returns a boolean.

The rationale is that you can name the methods in a way that allows others to understand what it is checking. For example, I do not understand why is the need to check for `input.length() < 6`. Maybe putting it in a method named 'hasInputIndex()` will make more sense. 

Another suggestion is that, perhaps checking length isn't the most suitable way to check for missing index? Maybe you can try string matching to match for numerics only.

bingsen0806 (commented):
Overall, a great job done! 👍 
I like how you did extensive JavaDocs comments for all your public methods and use reasonable names for your methods.

src/main/java/duke/gui/MainWindow.java
```diff
@@ -0,0 +1,81 @@
...
+
+    private final Image userImage = new Image(this.getClass().getResourceAsStream("/images/rottweiler.png"));
+    private final Image dukeImage = new Image(this.getClass().getResourceAsStream("/images/samoyed.png"));
+
+    /**
+     * Initialises a s
+     */
+    @FXML
+    public void initialize() {
+        String introMsg = "Hello, I'm Duke, your trusty doggo assistant!\nHere's what i can do:\n\n"
```
aceszhenwei: Perhaps this could be renamed to `introMessage` to avoid 'texting-style' spelling.

src/main/java/duke/TaskManager.java
```diff
@@ -0,0 +1,118 @@
...
+        return str.toString();
+    }
+
+    /**
+     * Returns a list of tasks with name matching the expr.
+     *
+     * @param expr the keyword used to search for matching tasks.
+     * @return list of tasks with names matching the expr.
+     */
+    public ArrayList<Task> findListOfMatchingTasks(String expr) {
```
aceszhenwei, line 107: Nice! I have seen that you have used a good habit throughout your code to let your names not just differentiate themselves, but also explain what they do!

src/main/java/duke/exceptions/UnknownCommandException.java
```diff
@@ -0,0 +1,10 @@
+package duke.exceptions;
+
+/**
+ * Exception thrown when user enters a command that duke.Duke has not been programmed to handle.
+ */
```
aceszhenwei, line 5: Well done on leaving javadocs for your exceptions! This is very helpful especially since these are custom exceptions.

aceszhenwei (commented):
Really well done! I found your code easy to read and follow. Apart from minor issues that @bingsen0806 may have mentioned, I think you are LGTM!

PR #81, opened 2022-01-22
[Sean Leong] iP

seanleong339:
# Duke
> One of the most dangerous forms of human error is forgetting what one is trying to achieve. - Paul Nitze [Source](https://www.brainyquote.com/quotes/paul_nitze_205927?src=t_forgetting)

Never forget anything, or at least anything you write in Duke! It's,
- In English
- Simple to learn
- ~FAST~ *VERY* FAST to use

All you have to do is,

1. Download it from [here](https://github.com/seanleong339/ip/releases/tag/A-Jar)
2. Add your tasks
3. Let it manage your tasks for you 🤪 

And it is **FREE OF CHARGE**!

Features:
- [x] 3 kinds of tasks
- [x] Reminder of tasks due daily
- [x] Date due!

Some basic commands are:
```
todo
deadline
event
today
bye
```

```java
public class Main {
    }
}
```


src/main/java/duke/commands/AddDeadline.java
```diff
@@ -0,0 +1,41 @@
...
+    /**
+     * Creates a new Deadline task and adds it to the list
+     * @param ui Ui to communicate with user
+     * @param list DukeList that stores Tasks
+     */
+    @Override
+    public void execute(Ui ui, DukeList list) {
+        Deadlines ans = new Deadlines(msg, date);
+        list.add(ans);
+    }
```
AAlghrairy: Any reason you did not delegate the process of adding tasks to the list to your DukeList class? Also, any reason you did not have just one "AddTask" command class, rather than one for each task type?

I've noticed this issue in several other places.

src/main/java/duke/commands/TryAgain.java
```diff
@@ -0,0 +1,20 @@
+package duke.commands;
+
+import duke.tasklist.DukeList;
+import duke.ui.Ui;
+
+public class TryAgain extends Command {
```
AAlghrairy, line 6: Perhaps add a header for this class (as it is a bit unclear what this class is for)?

src/main/java/duke/commands/TryAgain.java
```diff
@@ -0,0 +1,20 @@
...
+
+    @Override
+    public void execute(Ui ui, DukeList list) {
+    }
```
AAlghrairy: Any reason why this method body is empty? Consider adding a header comment to explain why.

src/main/java/duke/parser/DukeParser.java
```diff
@@ -0,0 +1,202 @@
...
+            else if(k[0].equals("E")){
+                Task j = new Events(k[2],k[3]);
+                if(k[1].equals("true")){
+                    j.mark();
+                }
+                ans.add(j);
+            }
+        }
+        return ans;
+    }
```
AAlghrairy: Perhaps delegate this process to the Storage class instead (since its purpose is to deal with File processes)?

src/main/java/duke/parser/DukeParser.java
```diff
@@ -0,0 +1,202 @@
...
+            if (arg.length == 1) {
+                throw new MissingArgumentException();
+            } else {
+                return new FindCommand(arg[1]);
+            }
+        } else {
+            throw new InvalidCommandException();
+        }
+
+    }
```
AAlghrairy: Any reason you did not use Enums to specify the type of each command (e.g. Creating an Enum class for task types) rather than repeatedly using .equals() command?

Also, any reason you did not use a switch statement here instead of repeatedly using "if" and "else if"?

src/main/java/duke/task/Task.java
```diff
@@ -0,0 +1,75 @@
+package duke.task;
+
+import java.time.LocalDate;
+
+public class Task {
+
+    private String s;
+    private boolean isDone;
+    private LocalDate date;
```
AAlghrairy: `private LocalDate date;`

Consider deleting this variable (as well as its respective getter and setter methods) since it is only utilized in Events and Deadline child classes (It is good practice to make sure that variables and methods in a parent class are only included if they will be shared across all the parent class' respective child classes).

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,25 @@
+package duke;
+
+import duke.commands.Command;
+import duke.parser.DukeParser;
+import duke.storage.Storage;
+import duke.tasklist.DukeList;
+import duke.ui.Ui;
+
+
+public class Duke {
```
AAlghrairy, line 11: Perhaps add a header comment for each of the classes?

I noticed this issue in the other class files too.

src/main/java/duke/tasklist/DukeList.java
```diff
@@ -0,0 +1,131 @@
...
+            storage.store(a);
+        } catch( IndexOutOfBoundsException e) {
+            System.out.println("\nDuke: Wrong index to delete! Use \"list\" to see the current tasks.\n");
+        }
+    }
+
+    /**
+     * Prints a list of tasks due on current date to the console
+     */
+    public void todayTask(){
```
AAlghrairy:
```suggestion
    public void getTodayTask() {
```

src/main/java/duke/tasklist/DukeList.java
```diff
@@ -0,0 +1,131 @@
...
+        LocalDate cur = LocalDate.now();
+        String day = "\nDuke: Here are the tasks due today\n";
+        boolean b = true;
+        for(Task t: a){
```
AAlghrairy:
```suggestion
        for(Task t: a) {
```

src/main/java/duke/test.java
```diff
@@ -0,0 +1,16 @@
+package duke;
+
+import java.io.IOException;
+import java.util.Scanner;
+
+public class test {
```
AAlghrairy:
```suggestion
public class Test {
```

src/main/java/duke/task/Task.java
```diff
@@ -0,0 +1,75 @@
...
+     * Returns a string representation of the task
+     * @return String representation of this task
+     */
+    public String show() {
+        if(isDone) {
+            return "[X] " + this.s;
+        } else {
+            return "[ ] " + this.s;
+        }
+    }
```
AAlghrairy: Perhaps override the "toString" method for processes where you need to show the instance variables? 
```suggestion
    @Override
    public String toString() {
        if(isDone) {
            return "[X] " + this.s;
        } else {
            return "[ ] " + this.s;
        }
    }
```
I noticed this in the other Task child classes too.

AAlghrairy (commented):
Overall, I found the code to be alright, although  I noted some OOP violations in certain areas (e.g. Parser class performing the action of reading Files, which should be a function in the Storage class). I also noted that there were no header comments in the Classes. Consider using Enums for the task types in lieu of using .equals(), if that makes sense.

src/main/java/duke/tasklist/DukeList.java
```diff
@@ -0,0 +1,131 @@
...
+import duke.task.Deadlines;
+import duke.task.Events;
+import duke.task.Task;
+
+import java.time.LocalDate;
+import java.util.List;
+
+public class DukeList {
+
+    private List<Task> a;
```
AAlghrairy: Perhaps rename this variable to be more descriptive about it's usage (e.g. "tasks)?

src/main/java/duke/tasklist/DukeList.java
```diff
@@ -0,0 +1,131 @@
...
+    /**
+     * Adds a task to the list, prints add message to the console.
+     * @param t Task to be added
+     */
+    public void add(Task t) {
+        a.add(t);
+        storage.store(a);
+        System.out.println("\nDuke: Got it. I've added this task:\n      "
+                + t.show() + "\n      Now you have "
+                + this.a.size() + " tasks in the list.\n");
```
AAlghrairy: Perhaps delegate the process of outputting messages to the UI class (as this could violate OOP principle)? I've noticed this issue in the other methods of this class as well.

AAlghrairy (commented):
_Additional feedback_

src/main/java/duke/commands/TryAgain.java
seanleong339: It was cause I needed to return a Command object regardless from the method in my Parser class, so this command is used when an invalid command is input by the user. Its a design flaw on my part, will think of ways to fix it.

src/main/java/duke/parser/DukeParser.java
seanleong339: Was not familiar with Enums previously, will be adding your suggestions in subsequent commits.

src/main/java/duke/tasklist/DukeList.java
seanleong339: Hi, really appreciate all the comments that you've left, I think they were really helpful for me to identify the general problems with my code design and how I should improve. Will probably take me some time to fix the code design issues but thanks a lot!

AAlghrairy: No problem. Glad it could help 👍

PR #83, opened 2022-01-22
[Tay Yan Han]/iP

tyanhan:
Saitama frees your mind of having to remember things you need to do. It's,
- text-based
- easy to learn
- ~~FAST~~ SUPER FAST to use

All you need to do is,

1. download it from [here](https://github.com/tyanhan/ip/releases)
2. double-click it.
3. add your tasks.
4. let Saitama manage your tasks for you! 👊 

And it is **FREE**!

Features:

- [x] Manage tasks
- [x] Manage deadlines
- [ ] Reminders (coming soon)

```java
        new Saitama(filePath).run();
}
```


KwanHW (commented):
Hi Yan Han!

Well done on your codestyle and naming of variables/functions. The naming was consistent and easy to follow.

Some nits you can do:
- Your compiled java classes are not ignored in the repository. You can do so by the steps below
1. Add this line `*.class` into your `.gitignore` file
2. [Untrack](https://www.codeblocq.com/2016/01/Untrack-files-already-added-to-git-repository-based-on-gitignore/) the existing `.class` files using the attached URL

src/main/java/saitama/parser/Parser.java
```diff
@@ -0,0 +1,192 @@
...
+            return prepareFind(splitCommand);
+        case "TODO":
+            //Fallthrough
+        case "DEADLINE":
+            //Fallthrough
+        case "EVENT":
+            //Fallthrough
+            return prepareAdd(splitCommand);
+        default:
+            throw new InvalidCommandException();
```
SethCKL: Really like the choice to abstract every single command into it's own function.

Makes the code very readable and easy to maintain going forward should one choose to change the functionality or add new commands.

LGTM! code is clean and follows the coding standards

src/main/java/saitama/tasks/TaskList.java
```diff
@@ -0,0 +1,114 @@
...
+    @Override
+    public String toString() {
+        int counter = 1;
+        String output = "";
+        for (Task task : taskList) {
+            output += counter + "." + task + "\n";
+            counter += 1;
+        }
+        return output.substring(0, output.length() - 1);
+    }
```
chriswengqi, line 108: LGTM! Would suggest that you use `StringBuilder` instead of string concatenation `+=` for better efficiency in large test cases.

src/main/java/saitama/parser/Parser.java
chriswengqi: Agree with Seth! Really liked the idea of abstracting every single command into its own function in its respective class. As a reader I feel comfortable in reading your code.

chriswengqi (approved):
LGTM! I learned a lot from reading your code :)

PR #97, opened 2022-01-23
[Jan Alfenson Tan] iP

Cyolune:
# Sana is here!
> "Sana is BEEEEG." - Sana ([source](https://www.youtube.com/channel/UCsUj0dszADCGbF3gNrQEuSQ))

Sana is here to help you keep track of things. She's
- ~text~-based
- pretty fast considering her size
- kinda cute ngl

All you need to do is,
1. download it from [here](https://github.com/Cyolune/ip/releases/tag/A-Jar)
2. run it on the command line
3. play!

And its **FREE**! :star_struck:

Features:
- [x] Managing tasks
- [x] Managing deadlines
- [ ] Reminders (coming soon)

If you are a Java programmer, you can use it to practice Java too! Here's the `main` method:
```java
public class Main {
    }
}
````

### CODE REVIEW

File: src/main/java/sana/Memory.java [hunk 1]
```diff
@@ -0,0 +1,103 @@
...
+        }
+        memFile = new File(MEM_PATH);
+        if (!memFile.exists()) {
+            try {
+                memFile.createNewFile();
+            } catch (IOException e) {
+                e.printStackTrace();
+            }
+        }
+        this.memFile = new File(MEM_PATH);
```
TypeDefinition: I noticed that the `this` keyword is not used when referring to `memFile` anywhere else in this function. Perhaps consider making this line consistent too?

src/main/java/sana/task/Task.java
```diff
@@ -0,0 +1,78 @@
...
+    }
+
+    /**
+     * Sets the sana.task to done or not
+     *
+     * @param done  sets the sana.task's completion
+     */
+    public void setDone(boolean done) {
+        this.isDone = done;
+    }
```
TypeDefinition, line 72: As per the [coding standard](https://se-education.org/guides/conventions/java/intermediate.html#naming), perhaps consider renaming `done` to `isDone`?

TypeDefinition (commented):
You've followed the coding standard very well! I struggled to find any mistakes at all! I also find the incorporation of Space Australian English into the bot's replies very cute. :australia: :smiley:

src/main/java/sana/Sana.java
```diff
@@ -58,29 +50,30 @@ public String doCommand(String userCommand) {
...
+            case "todo":
+            case "event":
+            case "deadline":
+                return new AddTask().executeCommand(parsedCmd, userTasks);
+            case "delete":
+                return new DeleteCommand().executeCommand(parsedCmd, userTasks);
+            case "find":
+                return new FindCommand().executeCommand(parsedCmd, userTasks);
+            default:
+                throw new UnknownCommandException();
```
CS2103Lerulatus: ——— For PR review ———
May I suggest the further use of OOP-styles -- package the case statements into the Command object itself. A `ByeCommand` object should contain the alias `"bye"` that trigger it's execution, for example.

Most commands would benefit from having multiple aliases that trigger their execution too, such as `bye, exit, quit` all being valid and recognized commands that executes the `ByeCommand` action. Writing all these aliases into a switch-case would prove troublesome, especially as the project scales and expands to include new commands and features, so packing them into their respective Command object would prove to be handy for future scaling.

Also, this switch-case block follows the recommended indentation, so that's good.

src/main/java/sana/Memory.java
```diff
@@ -0,0 +1,103 @@
+package sana;
+
+import java.io.File;
+import java.io.FileNotFoundException;
+import java.io.FileWriter;
+import java.io.IOException;
+import java.util.LinkedList;
+import java.util.Scanner;
```
CS2103Lerulatus, line 8: ——— For PR review ———
Nice and clear use of imports, even when `import java.io.*` is tempting to write in projects that extensively use io functionality.

PR #101, opened 2022-01-23
[Lai Mei Tin] iP

1001mei:
Duke frees you from the hassle of remembering all the tasks you need to do. It's
- text-based
- fast to learn
- ~~EASY~~ *SUPER EASY* to use

All you need to do is
1. download it from [here](https://github.com/1001mei/ip/releases)
2. run `java -jar duke.jar` command in your terminal to launch the application
3. add your tasks
4. let it keeps track of the tasks for you :grin:

And it is **FREE**!

Features:
- [x] Managing tasks
- [ ] Sort tasks based on deadline (coming soon) 
- [x] Find tasks using keyword(s)

## Extra
If you are a Java programmer, feel free to tweak the code and customize it to better suits your needs!
Here is a snippet from the `main` method.
```java
public class Launcher {
    }
}
```


src/main/java/duke/command/AddDeadlineTaskCommand.java
```diff
@@ -0,0 +1,44 @@
...
+import java.time.LocalDate;
+
+import duke.storage.Storage;
+import duke.task.TaskList;
+import duke.ui.Ui;
+
+/**
+ * Represents a command that when executed adds a deadline task to the task list.
+ */
+public class AddDeadlineTaskCommand extends Command {
```
janald99, line 12: I like how the names of your classes are super clear and straightforward 👍

src/main/java/duke/command/AddDeadlineTaskCommand.java
```diff
@@ -0,0 +1,44 @@
...
+
+    /**
+     * Class constructor.
+     * 
+     * @param description description of the deadline task.
+     * @param date deadline of the task expressed in date.
+     */
+    public AddDeadlineTaskCommand(String description, LocalDate date) {
+        this.description = description;
+        this.date = date;
```
janald99, line 24: Perhaps the Bot could integrate time into Deadline as well?

src/main/java/duke/command/AddDeadlineTaskCommand.java
```diff
@@ -0,0 +1,44 @@
...
+    }
+
+    /**
+     * Invokes <code>addDeadlineTask</code> method of <code>taskList</code> to add a deadline task.
```
janald99, line 28: I like your java doc comments a lot. They are very detailed 💯

src/main/java/duke/command/AddEventTaskCommand.java
```diff
@@ -0,0 +1,42 @@
...
+     * @param ui user interface of the application.
+     * @param taskList task list of the application.
+     * @param storage disk storage of the application.
+     */
+    @Override
+    public void execute(Ui ui, TaskList taskList, Storage storage) {
+        taskList.addEventTask(description, at);
+
+        int taskIndex = taskList.getNumberOfTasks() - 1;
+        ui.showMessage("Got it. I've added this task: ");
```
janald99, line 41: What if there is a Messages class that consists of the messages that you want to show?
Since "Got it. I've added this task: " may be printed multiple times, maybe you can abstract that into a class and call ui.showMessage(Messages.TASK_ADDED_MSG).
Just an alternative to consider! 😄 It may or may not be more of a hassle.

src/main/java/duke/command/ByeCommand.java
```diff
@@ -0,0 +1,37 @@
...
+    }
+
+    /**
+     * Checks if this command is a goodbye command.
+     * 
+     * @return true.
+     */
+    @Override
+    public boolean isBye() {
+        return true;
```
janald99, line 38: an alternative(that i saw from someone else's code) you can consider is:
```
return this instanceof ByeCommand;
```

in the abstract Command class.
That may help readers understand better, but it could be less efficient.
What do you think about this?

src/main/java/duke/command/ListTaskCommand.java
```diff
@@ -0,0 +1,32 @@
...
+        int numberOfTasks = taskList.getNumberOfTasks();
+
+        if (numberOfTasks == 0) {
+            ui.showMessage("Congratulations! There are no tasks in your list :)");
+            return;
+        }
+
+        ui.showMessage("Here are the tasks in your list:");
+        for (int i = 0; i < numberOfTasks; i++) {
+            ui.showMessage(String.format("%d. %s", i + 1, taskList.getDescriptionOfTaskAtIndex(i)));
```
janald99, line 31: perhaps the name of the method ```getDescriptionOfTaskAtIndex(i)``` is a little long? Although it explains very clearly what the method does. 🤔

janald99 (commented):
LGTM. The names of all variables and methods are super specific and detailed! The only trade-off for detailed names are that it may be too long, but it is worth it for readers to understand your code better, i believe 👍

src/main/java/duke/command/AddDeadlineTaskCommand.java
1001mei: I agree with you. Will implement this in the future.

src/main/java/duke/command/AddDeadlineTaskCommand.java
lkldev: Agree with @/janald99, great class names and easy to understand! 👍

src/main/java/duke/command/SetupCommand.java
```diff
@@ -0,0 +1,26 @@
...
+
+import duke.exception.DukeException;
+import duke.storage.Storage;
+import duke.task.TaskList;
+import duke.ui.Ui;
+
+/**
+ * Represents a command that when executed prepares the application for getting commands from user.
+ */
+public class SetupCommand extends Command {
```
lkldev, line 11: I like how you have a SetupCommand to load the data from storage 👍 It makes the code neater without having the main to initialize the storage before running.

src/main/java/duke/parser/Parser.java
```diff
@@ -0,0 +1,71 @@
...
+import duke.command.ListTaskCommand;
+import duke.command.MarkTaskCommand;
+import duke.command.UnmarkTaskCommand;
+import duke.exception.DukeException;
+
+/**
+ * This class acts as a parser that parses user inputs into specific commands.
+ */
+public class Parser {
+    private static final int OFFSET = 1;
```
lkldev: Great use of constants to replace magic numbers 👍 However, I feel that maybe `ARRAY_INDEX_OFFSET` might be a more meaningful name?

src/test/java/duke/parser/ParserTest.java
```diff
@@ -0,0 +1,50 @@
+package duke.parser;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.fail;
+
+import org.junit.jupiter.api.Test;
+
+import duke.exception.DukeException;
+
+public class ParserTest {
```
lkldev, line 10: I like how you break all the test cases into different methods 👍 I would have created one method and tested everything related to the feature under testing in it. E.g. `parse_invalidMarkCommand` to test both `parse_invalidMarkCommandFormat` and `parse_incompleteMarkCommand`. But your way is definitely more comprehensive 👍

src/main/java/duke/task/TaskList.java
```diff
@@ -0,0 +1,176 @@
...
+     * Populates this task list with the task data stored in disk.
+     * 
+     * @param data data read from storage.
+     * @throws DukeException upon encountering incorrect format of data record.
+     */
+    public void populateWith(String[] data) throws DukeException {
+        try {
+            numberOfTasks = data.length;
+            for (int i = 0; i < numberOfTasks; i++) {
+                String tokens[] = data[i].split(",");
```
lkldev: Based on the java coding standard, the array specifiers must be attached to the type not the variable, e.g. 
```suggestion
                String[] tokens = data[i].split(",");
```

src/main/java/duke/ui/Ui.java
```diff
@@ -0,0 +1,48 @@
+package duke.ui;
+
+import java.io.PrintStream;
+
```
lkldev: Seems like there is an extra newline here between the same `java.*` import group.

src/main/java/duke/ui/Ui.java
```diff
@@ -0,0 +1,48 @@
+package duke.ui;
+
+import java.io.PrintStream;
+
+import java.util.Scanner;
+
+public class Ui {
```
lkldev, line 8: Based on the java coding standard, it will be good if you add comments for all public classes and methods.

src/main/java/duke/command/AddEventTaskCommand.java
1001mei: This is a good idea! May add this class in the future.

src/main/java/duke/command/ByeCommand.java
1001mei: I think the alternative is fine also, but I would still prefer overriding it in ByeCommand class because Command class knows nothing about its subclass, and I personally feel that writing code in superclass that refers to subclass is a bit weird. Just personal preferences I guess 😀

lkldev (commented):
Overall, Great work! Love the descriptive method names you provided. Other than that, below are some nits I have picked out using the help of checkstyle, you may want to integrate it into your project to help you spot the coding standard related issues 👍

src/main/java/duke/command/ByeCommand.java
janald99: that makes a very good point. I am convinced this is better :)

src/main/java/duke/parser/Parser.java
1001mei: I agree, thanks for the suggestion!

src/main/java/duke/task/TaskList.java
1001mei: Thanks for pointing out this coding standard violation, will resolve it in coming commits.

PR #102, opened 2022-01-23
[Gan Ting, Nicolle] iP

nicollegann:
- text-based
- easy to learn

All you need to do is,
1. download it from [here](https://github.com/nicollegann/ip/releases/tag/A-Jar).
2. double-click it.
3. add your tasks.

And it is **FREE**!

Features:
- [X] Managing tasks (to-dos, deadlines, events)
- [ ] Reminders (coming soon)
##
```java
public class Main {
    }
}
```


src/main/java/duke/command/AddCommand.java
```diff
@@ -0,0 +1,49 @@
...
+        taskList.add(this.task);
+        ui.printOutput("Got it! I've added this task:\n  "
+                + ui.showIndent() + this.task + "\n"
+                + ui.showIndent() + taskList.getListStatus());
+
+        storage.saveToHardDisk(taskList);
+    }
+
+    /**
+     * Returns whether this is an exit command.
```
takufunkai: Might be clearer to explain explicitly what exit and non-exit commands will return respectively, for example `Returns false if and only if the Command is an ExitCommand`.

src/main/java/duke/command/DeleteCommand.java
```diff
@@ -0,0 +1,52 @@
...
+     * Class constructor.
+     *
+     * @param taskNum The index of the task to be deleted.
+     */
+    public DeleteCommand(int taskNum) {
+        this.taskNum = taskNum;
+    }
+
+    /**
+     * Execute the delete command, print confirmation message and saves the updated task list to file.
```
takufunkai: I think it should be executes, prints and saves, and I think it was repeated in several other java-docs you wrote for `Command::execute`

src/main/java/duke/command/ExitCommand.java
```diff
@@ -0,0 +1,33 @@
...
+import duke.util.Storage;
+import duke.util.TaskList;
+import duke.util.Ui;
+
+/**
+ * Represents the exit command to end the program.
+ */
+public class ExitCommand extends Command {
+
+    public ExitCommand() {}
```
takufunkai: Might be cleaner to leave it as [default constructor](https://beginnersbook.com/2014/01/default-constructor-java-example/) instead? Only scenario I can imagine to have this line is if we want to keep our constructor `private` so that other classes will not be able to call this constructor.

src/main/java/duke/command/FindCommand.java
```diff
@@ -0,0 +1,41 @@
...
+     *
+     * @param keyword The keyword used to find tasks.
+     */
+    public FindCommand(String keyword) {
+        this.keyword = keyword;
+    }
+
+    /**
+     * Executes the find command to find tasks containing the keyword and print them.
+     * @param taskList The current list of tasks.
```
takufunkai: We're told to leave a one-line buffer between parameters and the description

src/main/java/duke/util/Parser.java
```diff
@@ -0,0 +1,144 @@
...
+                        return new FindCommand(description);
+                    }
+                }
+
+            case "bye":
+                return new ExitCommand();
+
+            default:
+                throw new DukeException("Oh no! I don't understand what that means...");
+            }
```
takufunkai: Switch-case statements should not have a line-break between them.

src/main/java/duke/command/AddCommand.java
```diff
@@ -0,0 +1,49 @@
...
+     * @param storage The storage of the program.
+     * @throws DukeException If updated task list cannot be saved to file.
+     */
+    public void execute(TaskList taskList, Ui ui, Storage storage) throws DukeException {
+        taskList.add(this.task);
+        ui.printOutput("Got it! I've added this task:\n  "
+                + ui.showIndent() + this.task + "\n"
+                + ui.showIndent() + taskList.getListStatus());
+
+        storage.saveToHardDisk(taskList);
```
takufunkai: I really like that your code is segmented into logical blocks of work!

takufunkai (commented):
Your program abides by most if not all the Java coding standards, and your Java Docs are also well written. You kept all your fields private, and also segmented some of your code into logical blocks of work. Your methods also flow logically to some degree. I think you made this very easy for me 😅😅💯

src/main/java/duke/command/DeleteCommand.java
nicollegann: Thanks for pointing out these! Your comments have been really helpful 😄

src/main/java/duke/command/DeleteCommand.java
```diff
@@ -0,0 +1,52 @@
...
+     */
+    public void execute(TaskList taskList, Ui ui, Storage storage) throws DukeException {
+        int currentTotalTasks = taskList.getTotalTasks();
+        if (taskNum <= currentTotalTasks && taskNum > 0) {
+            Task taskDeleted = taskList.delete(this.taskNum);
+            ui.printOutput("Okay, I've deleted this task:\n  " + ui.showIndent()
+                    + taskDeleted + "\n" + ui.showIndent() + taskList.getListStatus());
+            storage.saveToHardDisk(taskList);
+        } else {
+            throw new DukeException("Invalid task number. There are " + currentTotalTasks + " in the list.");
```
petermonky: I like how detailed your `DukeException` messages are. I can definitely see it helping the user.

src/main/java/duke/command/InvalidCommand.java
```diff
@@ -0,0 +1,32 @@
+package duke.command;
+
+import duke.util.Storage;
+import duke.util.TaskList;
+import duke.util.Ui;
+
+/**
+ * Represents an invalid command.
+ */
+public class InvalidCommand extends Command {
```
petermonky, line 10: I like how you have a separate class for an invalid command; I would not have thought to handle invalid commands like this.

src/main/java/duke/util/Parser.java
```diff
@@ -0,0 +1,154 @@
...
+import duke.task.Event;
+import duke.task.Task;
+import duke.task.Todo;
+
+/**
+ * Represents the parse function of the program. Deals with parsing user input.
+ */
+public class Parser {
+
+    public Parser() {}
```
petermonky: I don't think this constructor is necessary; if I'm not wrong, Java implicitly provides the default constructor for all classes.

src/main/java/duke/util/Parser.java
```diff
@@ -0,0 +1,154 @@
...
+            case "bye":
+                return new ExitCommand();
+            default:
+                throw new DukeException("Oh no! I don't understand what that means...");
+            }
+        } catch (DateTimeParseException e) {
+            throw new DukeException("Invalid date/time format.\n    "
+                    + "Please use the following format: yyyy-mm-dd HH:mm");
+        }
+        return new InvalidCommand();
```
petermonky: Not sure if I'm understanding this right, but is this line ever reached? I feel like the default clause of the switch-case statement already handles invalid commands.

src/main/java/duke/util/Parser.java
```diff
@@ -0,0 +1,154 @@
...
+        }
+    }
+
+    /**
+     * Returns a boolean indicating whether there is a given keyword provided in the task description.
+     *
+     * @param descrArr The description split by the keyword.
+     * @param keyword The keyword to be found in the description.
+     * @param taskType The type of the task (e.g. deadline, event)
+     * @return True if keyword can be found, false otherwise.
```
petermonky: Is this `@return` descriptor accurate? The method only seems to return `true`.

src/main/java/duke/util/Parser.java
```diff
@@ -0,0 +1,154 @@
...
+        } else {
+            return true;
+        }
+    }
+
+    /**
+     * Returns LocalDateTime.
+     *
+     * @param dt The string of format yyyy-MM-dd HH:mm
+     * @return LocalDateTime.
```
petermonky, line 180: Perhaps could elaborate on what `LocalDateTime` it is? i.e. the parsed `LocalDateTime`?

src/main/java/duke/util/Storage.java
```diff
@@ -0,0 +1,110 @@
...
+        try {
+            if (!folder.exists()) {
+                folder.mkdir();
+            } else {
+                File file = new File(filePath + fileName);
+                file.createNewFile();
+            }
+        } catch (IOException e) {
+            System.out.println("Unable to create file to save tasks.");
+        }
```
petermonky, line 43: I think this part can be modified to make `filePath` more accurate; as of now `filePath` is being treated more like the path of the directory that the file is located in. Perhaps could make use of `getParentFile()` and `mkdirs()`? i.e. from

```java
File folder = new File(filePath);
try {
    if (!folder.exists()) {
        folder.mkdir();
    } else {
        File file = new File(filePath + fileName);
        file.createNewFile();
    }
} catch (IOException e) {
    System.out.println("Unable to create file to save tasks.");
}
```
to
```java
File file = new File(filePath);
try {
    if (!file.exists()) {
        file.getParentFile().mkdirs();
        file.createNewFile();
    }
} catch (IOException e) {
    System.out.println("Unable to create file to save tasks.");
}
```

src/main/java/duke/util/Storage.java
```diff
@@ -0,0 +1,110 @@
...
+     */
+    public void saveToHardDisk(TaskList taskList) throws DukeException {
+        try {
+            String formattedTaskList = taskList.formatListForSaving();
+            FileWriter fw = new FileWriter(filePath + fileName);
+            fw.write(formattedTaskList);
+            fw.close();
+        } catch (IOException e) {
+            throw new DukeException("Unable to save changes to hard disk :(");
+        }
```
petermonky, line 113: Shouldn't `fw.close()` be in the `finally` block? Since control wouldn't return to the `try` block, should an error occur before `fw.close()`.

src/main/java/duke/util/TaskList.java
```diff
@@ -0,0 +1,161 @@
...
+
+    /**
+     * Returns the list of matching tasks. If no task matches keyword, return not found message.
+     *
+     * @param keyword The keyword used to search for tasks.
+     * @return String representation of matching tasks.
+     */
+    public String findTasksWithKeyword(String keyword) {
+        String list = "Here are the matching tasks in your list:";
+        int index = 1;
```
petermonky: Is `index` used in this method, other than when it is incremented in line 152?

src/main/java/duke/util/TaskList.java
```diff
@@ -0,0 +1,161 @@
...
+        return taskToUnmark;
+    }
+
+    /**
+     * Returns the list of matching tasks. If no task matches keyword, return not found message.
+     *
+     * @param keyword The keyword used to search for tasks.
+     * @return String representation of matching tasks.
+     */
+    public String findTasksWithKeyword(String keyword) {
```
petermonky: I feel like the name of this method is a bit misleading; one might assume the method returns a `TaskList`. Perhaps the method name can be changed to better reflect the return type?

src/main/java/duke/util/Ui.java
```diff
@@ -0,0 +1,85 @@
...
+    }
+
+    /**
+     * Returns the default indentation of output.
+     *
+     * @return Indentation of four spaces.
+     */
+    public String showIndent() {
+        return indent;
+    }
```
petermonky, line 19: I like how you abstracted away the string constants to individual methods!

petermonky (commented):
Overall, a very nicely structured project; clear use of encapsulation here and there. Other than some minor code quality improvements, a solid piece of work.

PR #106, opened 2022-01-23
[Ng Jun Kang] iP

ngjunkang:
**BCB** allows fellow programmers like us to practise the habit of keeping "list of goals" to "stay on track". When programming, we often lose track of our goals and the time, and we forget what we were trying to achieve at the start. BCB is:
- Perfect for programmers
- Similar to our working environment (CLI)
- Fast with hardly any downtime

#### All you need to do is:
1. Download the jar file from the release page [here](https://github.com/ngjunkang/ip/releases).
2. Run it with the `java -jar` command, e.g. `java -jar baron.jar`.
3. Add your tasks.
4. Let BCB manage all your tasks.

And most importantly, unlike the applications in the market, it is 💯% **FREE**!!

#### Features:
- [x] Add ToDo
- [x] Add Event
- [x] Add Deadline
- [x] List all tasks
- [x] Mark task
- [x] Un-mark task
- [x] Delete task

```java
    new Baron("data/baron.txt").start();
}
```


src/main/java/baron/util/TextUi.java
```diff
@@ -0,0 +1,29 @@
+package baron.util;
+
+import baron.messages.Messages;
+import baron.tasks.TaskManager;
+
+public class TextUi {
```
tandeshao, line 9: I like how you have abstracted out the UI from Duke main logic 😄 !

src/main/java/baron/messages/Messages.java
```diff
@@ -8,15 +8,20 @@
...
-    public static final String MESSAGE_FILE_NOT_FOUND = "Data file is not found.\n";
-    public static final String MESSAGE_INVALID_FILE_FORMAT = "Data file format is invalid.\n";
-    public static final String MESSAGE_FILE_CREATION_FAIL = "Data file cannot be created.\n";
-    public static final String MESSAGE_FILE_WRITE_FAIL = "Data file cannot be written to.\n";
+    public static final String MESSAGE_NO_TASK = "You have no tasks.";
+    public static final String MESSAGE_FILE_NOT_FOUND = "Data file is not found.";
+    public static final String MESSAGE_INVALID_FILE_FORMAT = "Data file format is invalid.";
+    public static final String MESSAGE_FILE_CREATION_FAIL = "Data file cannot be created.";
+    public static final String MESSAGE_FILE_WRITE_FAIL = "Data file cannot be written to.";
     public static final String MESSAGE_DATE_TIME_FORMAT_INVALID = "The date/time should be in the following format: d/M/yyyy HH:mm";
```
tandeshao: I didn't know you could name all the default messages as an attribute, I'm impressed!

src/main/java/baron/commands/CommandManager.java
```diff
@@ -62,6 +62,8 @@ public Command parseCommand(String fullCommand) {
                 return new AddTaskCommand(this.taskManager, this.storage, TaskType.EVENT, commandArg);
             case "delete":
                 return new DeleteTaskCommand(this.taskManager, this.storage, commandArg);
+            case "find":
+                return new FindCommand(this.taskManager, commandArg);
             default:
```
tandeshao: In Intellij they allow enhanced switch-case statements, you might want to change your switch-case control flow for better readability. An example of enhanced switch-case statements:

```java
switch (itemCode) {
    case 001 -> System.out.println("It's a laptop!");
    case 002 -> System.out.println("It's a desktop!");
    case 003,004 -> System.out.println("It's a mobile phone!");
}
```

src/main/java/baron/tasks/Task.java
```diff
@@ -29,10 +54,21 @@ public boolean unmark() {
...
         return "[" + this.getStatusIcon() + "] " + description;
     }
 
+    /**
+     * Returns the encoded string to be stored in the storage using the specified delimiter.
+     *
+     * @param delimiter the string to be inserted in between various information of this task.
+     * @return the encoded string to be stored in the storage using the specified delimiter.
+     */
     public String toSaveString(String delimiter) {
```
tandeshao, line 80: I believe ``saveString`` is sufficient.

src/main/java/baron/commands/UnmarkTaskCommand.java
```diff
@@ -5,22 +5,41 @@
...
 import baron.util.Storage;
 
+/**
+ * Represents a unmark task command that unmarks the task during execution.
+ */
 public class UnmarkTaskCommand extends Command {
     private final TaskManager taskManager;
     private final Storage storage;
-    private final String commandArgs;
+    private final String commandArg;
```
tandeshao, line 14: I like how you change the name of ``commandArgs`` to ``commandArg`` since ``commandArg`` only represents a single command

src/main/java/baron/commands/CommandManager.java
```diff
@@ -33,20 +33,22 @@ public Command parseCommand(String fullCommand) {
...
+            case "deadline":
+                return new AddTaskCommand(this.taskManager, this.storage, TaskType.DEADLINE, commandArg);
+            case "event":
+                return new AddTaskCommand(this.taskManager, this.storage, TaskType.EVENT, commandArg);
+            case "delete":
+                return new DeleteTaskCommand(this.taskManager, this.storage, commandArg);
+            case "find":
+                return new FindCommand(this.taskManager, commandArg);
+            default:
+                return new InvalidCommand();
```
tandeshao: Consider using enhanced switch-case statements.

tandeshao (commented):
Code looks clean and succinct 😄 ! LGTM

src/main/java/baron/commands/CommandManager.java
ngjunkang: Hi De Shao, I didn't know there's such a syntax in Java, looks similar to lambdas. Good to know! And I will definitely use in the future. However, I won't be applying to my code, because I am not sure it is compliant to the coding standard and the Java version (JDK 11, it's a JDK 12-13 feature, I think) of this course. Thank you for the suggestion though.

src/main/java/baron/tasks/Task.java
ngjunkang: Hi De Shao, thank you for the suggestion, but I feel that as this method doesn't save string, so `saveString` would be misleading (it may mean that this method saves the string somewhere, `save` as in the verb). In fact, it just returns the encoded string that is to be saved somewhere by other parts of my code.

src/main/java/baron/util/TextUi.java
ngjunkang: Thank you! :)

src/main/java/baron/messages/Messages.java
ngjunkang: Hi De Shao, my reason for this design is actually just for me to easily change the outputs, instead of finding different Java files to fix the outputs. For example, if I want to change to a more kids-friendly tone. Glad that you are impressed!

src/main/java/baron/commands/UnmarkTaskCommand.java
ngjunkang: Thank you! :)

src/main/java/baron/tasks/TaskManager.java
```diff
@@ -0,0 +1,271 @@
...
+            throw new BaronException(Message.generateTaskIndexOutOfBoundMessage(this.getTaskCount()));
+        }
+    }
+
+    /**
+     * Filters the task list in this {@code TaskManager} by the specified {@code keyword}.
+     *
+     * @param keyword the keyword to filter the task list by.
+     * @return the filtered task list.
+     */
```
lawwm, line 212: Good use of fp.

src/main/java/baron/tasks/TaskManager.java
```diff
@@ -0,0 +1,271 @@
...
+     * {@code unmarkTask}).
+     * <p>
+     * To use this method, the instance variable {@code previousTaskList} must be set
+     * to the state of {@code taskList} using {@code this.previousTaskList = this.getAllTasks();}.
+     * {@code previousTaskList} can be cleared using {@code commitChanges} below.
+     *
+     * @see #taskList
+     * @see #previousTaskList
+     * @see #commitChanges()
+     */
```
lawwm, line 235: Seems like it depends on client using it correctly (after an edit operation). Maybe can return boolean value depending on if there is a previous change to revert to?

src/main/java/baron/tasks/TaskManager.java
```diff
@@ -0,0 +1,271 @@
...
+    }
+
+    /**
+     * Un-marks the task that is at the specified position (1-indexed).
+     *
+     * @param index the index of the task (1-indexed).
+     * @return true if the task is un-marked successfully.
+     * @throws BaronException If there is no task, the specified task was not marked,
+     *                        or the index is out of bound.
+     */
```
lawwm, line 171: Maybe you can use guard clauses to make happy path more prominent?

lawwm (commented):
~~Removed.~~

src/main/java/baron/commands/DeleteTaskCommand.java
```diff
@@ -0,0 +1,57 @@
...
+                this.taskManager.revertChanges();
+                throw e;
+            }
+            this.taskManager.commitChanges();
+            return Message.MESSAGE_DELETE_SUCCESS + deletedTask.toString() + "\n"
+                    + Message.generateNoOfTasksMessage(this.taskManager.getTaskCount());
+        } catch (BaronException e) {
+            return e.toString();
+        }
+    }
```
lawwm, line 56: Is it possible to refactor the nested try catch statements into one try catch statement for better readability?

src/main/java/baron/tasks/TaskManager.java
```diff
@@ -0,0 +1,271 @@
...
+     * @param keyword the keyword to filter the task list by.
+     * @return the filtered task list.
+     */
+    public ArrayList<Task> filterDescByKeyword(String keyword) {
+        ArrayList<Task> filteredTaskList = new ArrayList<>();
+        this.taskList.forEach((task -> {
+            if (task.description.contains(keyword)) {
+                filteredTaskList.add(task);
+            }
+        }));
```
lawwm: Good use of fp.

src/main/java/baron/tasks/TaskManager.java
```diff
@@ -0,0 +1,271 @@
...
+     * @see #taskList
+     * @see #previousTaskList
+     * @see #commitChanges()
+     */
+    public void revertChanges() {
+        if (this.previousTaskList != null) {
+            this.taskList = new ArrayList<>(this.previousTaskList);
+            this.previousTaskList = null;
+        }
+    }
```
lawwm: Seems like it depends on client using it correctly (after an edit operation). Maybe can return boolean value depending on if there is a previous change to revert to?

src/main/java/baron/tasks/TaskManager.java
```diff
@@ -0,0 +1,271 @@
...
+        } else if (index > 0 && index <= this.getTaskCount()) {
+            if (this.taskList.get(index - 1).mark()) {
+                return true;
+            } else {
+                throw new BaronException("The task is already marked.");
+            }
+        } else {
+            throw new BaronException(Message.generateTaskIndexOutOfBoundMessage(this.getTaskCount()));
+        }
+    }
```
lawwm: Maybe you can use guard clauses to make happy path more prominent? This can be applied for other methods in this class as well.

lawwm (commented):
Looks good, didn't notice coding standard violation. Sorry for reposting, I made a mistake on first review.

src/main/java/baron/tasks/TaskManager.java
ngjunkang: Hi, can you provide me with an example of a guard clause? :)

Jacky142857 (commented):
This is almost impeccable! Very good use of constants and JavaDoc you wrote is the best I've seen. You used {@code } to highlight the keyword in javadoc and @see to guide us to other descriptions. I think some nested if statement and try statement can be written in parallel to increase readability. You did quite a good job.

PR #108, opened 2022-01-23
[Kwan Hao Wei] iP

KwanHW:
# Waldo

> Can't find me if I'm on Quarantine Order!

Waldo frees your mind of having to remember things you need to do. It's,
- text-based
- easy to learn
- ~~FAST~~ SUPER FAST to use

All you need to do is,
1.  download it from [here](https://github.com/KwanHW/ip/releases).
2.  double-click it.
3.  add your tasks.
4.  let it manage your tasks for you 😉

And it is **FREE**!

Features:
- [x]     Managing tasks
- [x]     Managing deadlines
- [ ]     Reminders (coming soon)

<hr>

```java
        new Duke();
        Application.launch(Duke.class, args);
    }
```


src/test/java/duke/task/TaskStoreTest.java
```diff
@@ -0,0 +1,44 @@
...
+        assertTrue(new TaskStore().getIsEmpty(), "Failed: Task Store not empty when initialised");
+    }
+
+    @Test
+    public void noTasksOnDate() {
+        TaskStore ts = new TaskStore();
+        LocalDate date = LocalDate.parse("2022-01-31");
+        ts.addTask(new Deadline("Complete Assignment", date));
+        ts.addTask(new Event("Project Meeting", date));
+        assertEquals("You don't have any tasks on Jan 29 2022", ts.getTasksOn(LocalDate.parse("2022-01-29")), "Failed: TaskStore contains tasks that do not fall on that date");
```
LapisRaider, line 22: Line length longer than 120 characters, might want to warp it.

LapisRaider: Line length longer than 120 characters, might want to wrap it.

src/test/java/duke/task/TaskStoreTest.java
```diff
@@ -0,0 +1,44 @@
...
+    }
+
+    @Test
+    public void noTasksOnDate_noTimeables() {
+        TaskStore ts = new TaskStore();
+        LocalDate date = LocalDate.parse("2022-01-31");
+        ts.addTask(new Todo("Complete Assignment"));
+        ts.addTask(new Todo("Project Meeting"));
+        assertEquals("You don't have any tasks on Jan 29 2022", ts.getTasksOn(LocalDate.parse("2022-01-29")), "Failed: TaskStore only contains non-timeable instances.");
```
LapisRaider: Line length longer than 120 characters, might want to wrap it.

src/test/java/duke/task/TaskStoreTest.java
```diff
@@ -0,0 +1,44 @@
...
+    }
+
+    @Test
+    public void tasksOnDate() {
+        TaskStore ts = new TaskStore();
+        LocalDate date = LocalDate.parse("2022-01-29");
+        ts.addTask(new Deadline("Complete Assignment", date));
+        ts.addTask(new Event("Project Meeting", date));
+        String expected = "Here are your tasks on Jan 29 2022\n" + ts.getTask(0).toString() + "\n" + ts.getTask(1).toString() + "\n";
```
LapisRaider: Line length longer than 120 characters, might want to wrap it. There might be other lines longer than 120 characters that I couldn't catch. 

IntelliJ has this formatting option that will help you wrap the lines longer than a certain amount of characters (you can set the limit in settings). They have a bunch of other settings too, definitely a time saver instead of manually scanning through every line of the codes!

src/main/java/duke/task/Task.java
```diff
@@ -0,0 +1,59 @@
...
+     */
+    public String writeToFile() {
+        String isDone = this.isDone ? "1" : "0";
+        return String.join(" | ", this.taskType.toString(), isDone, this.description);
+    }
+
+    @Override
+    public String toString() {
+        return String.format("[%s][%s] %s", this.taskType.toString(), this.getStatusIcon(), this.description);
+    }
```
LapisRaider: The usage of `this.` keyword. It is recommended to use the `this` keyword only when a field is shadowed by a method or constructor parameter.

image

src/main/java/duke/task/Event.java
```diff
@@ -0,0 +1,55 @@
+package duke.task;
+
+import java.time.LocalDate;
+import java.time.format.DateTimeFormatter;
+
+/**
+ * Represents an event that is happening on a certain date.
+ */
+public class Event extends Task implements Timeable {
+    protected LocalDate at;
```
LapisRaider: Might want to change this to private instead of protected. For java, unlike C++ where protected makes the variable inaccessible to other classes besides the derived classes, in java, as long as a class is in the same package, they can access the protected variable.

If that is not your intention, you may want to change to private since Event class is only using this variable (correct me if I'm wrong).

src/main/java/duke/task/Deadline.java
```diff
@@ -0,0 +1,54 @@
+package duke.task;
+
+import java.time.LocalDate;
+import java.time.format.DateTimeFormatter;
+
+/**
+ * Represents a task that has to be completed before a certain date.
+ */
+public class Deadline extends Task implements Timeable {
+    protected LocalDate by;
```
LapisRaider: Might want to change this to private instead of protected. For java, unlike C++ where protected makes the variable inaccessible to other classes besides the derived classes, in java, as long as a class is in the same package, they can access the protected variable.

If that is not your intention, you may want to change to private since Deadline class is only using this variable (correct me if I'm wrong).

src/main/java/duke/io/Storage.java
```diff
@@ -0,0 +1,90 @@
+package duke.io;
+
+import duke.task.*;
```
LapisRaider: Might want to list out the imports explicitly instead of using wildcard imports since it's not recommended.

LapisRaider (commented):
Looks great! Just a few nits to fix.

TLDR:
- `this` keyword is recommended being used only when a field is shadowed by a method or constructor parameter.
- Some of your lines might be longer than 120 characters, IntelliJ can help you format.
- Protected might want to change to private. Protected keyword in Java makes the variable/function accessible for any class in the same package.
- Might want to change wildcard imports to listing explicitly instead.

src/main/java/duke/task/Event.java
KwanHW: Thank you for catching that error. I may have gotten it mixed up with experience in previous languages (like C#). 

I have resolved the other issues similar as the one mentioned

PR #118, opened 2022-01-23
[Ong Han Yang] iP

FestiveCat:
# ~Duke~ ***Fluffers***
## Added features so far:
- Level-1 to Level-9
- 1000% more [Cats](https://en.wikipedia.org/wiki/Cat) :smile_cat:
> to *meow*, or not to *meow*?
## TODO:
1. A-JUnit
2. Week 4 additions
## **Main method** from `Fluffers::main`
```ruby
    public static void main(String[] args) throws SaveFileModifiedException {
        Fluffers f = new Fluffers();

        System.out.println(f.greet());

        Scanner sc = new Scanner(System.in);

        while (f.isAwake) {
            String input = sc.nextLine();
            System.out.println(f.feedCommandAndReply(input));
        }
    }
```
## Task List
- [x] *Cats*
- [ ] JUnit tests
- [ ] Week 4 tasks


src/main/java/chatbots/Chatbot.java
```diff
@@ -0,0 +1,49 @@
...
+ */
+public abstract class Chatbot {
+    /** Whether the chatbot is awake or asleep. */
+    protected boolean isAwake;
+
+    /**
+     * Constructs a chatbot. Initial state of the chatbot is that it is asleep.
+     */
+    protected Chatbot() {
+        this.isAwake = false;
```
tyanhan, line 19: It is great that you made your booleans sound like booleans.

src/main/java/chatbots/Chatbot.java
```diff
@@ -0,0 +1,49 @@
...
+    }
+
+    /**
+     * Feeds a command and gives a reply based on the command.
+     *
+     * @param input the command as input.
+     * @return the reply based on the command.
+     */
+    public abstract String feedCommandAndReply(String input);
```
tyanhan: Would it be better to rename this function to getReply() or something else shorter instead?

src/main/java/date/time/DateTimeParser.java
```diff
@@ -0,0 +1,141 @@
...
+        if (input.length() != 16) {
+            throw INVALID_DATE_TIME_FORMAT;
+        }
+        // check for separators and numbers
+        // 0123456789012345
+        // yyyy-mm-dd xx:xx
+        // if wrong format, return false
+        // Any "magic numbers" seen from this point onwards refer to the indexes of the specific
+        // string to check, as shown above in line 80.
+        for (int i = 0; i < input.length(); i++) {
```
tyanhan: Would using switch statements help to reduce deep nesting and make your code more efficient?

src/main/java/date/time/DateTimeParser.java
```diff
@@ -0,0 +1,141 @@
...
+        int mins = Integer.parseInt((input.substring(14, 16)));
+        boolean isValidHour = hours >= 0 && hours < 24;
+        boolean isValidMin = mins >= 0 && mins < 60;
+        if (!isValidHour || !isValidMin) {
+            throw new InvalidTimeException(
+                    String.format("There is no %d-th hour, %d-th minute!", hours, mins));
+        }
+        return true;
+    }
+}
```
tyanhan, line 193: Perhaps you may want to use LocalDate::parse to help you check for valid date format instead of doing it manually yourself, but at the same time I think it is great that your customised exceptions allow for more intuitive error messages.

src/main/java/instructions/Instruction.java
```diff
@@ -0,0 +1,75 @@
...
+        String[] split = input.split(" ", 2);
+        String instType = split[0];
+        switch (instType) {
+        case "list":
+            return DisplayListInst.of();
+        case "hi":
+            return HelloInst.of();
+        case "bye":
+            return ExitInst.of();
+        case "mark":
```
tyanhan: Not sure if it is necessary, but I believe you might have to add //Fallthrough for switch cases without break statement.

tyanhan (commented):
Great job overall! I did not notice any jarring coding standard errors, and the variable/function names were all well chosen as well. Keep up the good work.

src/main/java/instructions/Instruction.java
FestiveCat: oh good point, I'll be sure to add it, thanks!

src/main/java/date/time/DateTimeParser.java
FestiveCat: Hmmm, what do you mean?
I thought about maybe abstracting each part away into individual helper functions/methods but I'm not too sure how I would use a switch statement here

src/main/java/chatbots/Chatbot.java
FestiveCat: Hmm yeah I think that's a good point, thanks!

src/main/java/date/time/DateTimeParser.java
tyanhan: I was thinking of using switch(i) to check the value of the index instead of multiple if/else statements. This will also help to reduce the amount of nesting and make the code look nicer. But on a second look I guess the amount of nesting wasn't that bad. 

I don't think it is necessary to abstract each individual part as this check is a pretty simple check. Rather than that, 1 suggestion I have(but may not be the most optimal) would be to use String::split and then doing the necessary checks by 1) checking the appropriate lengths of each split, and 2) checking if the splits that are supposed to be digits are digits. The con of this is that it might lengthen your code, but it avoids magic numbers. Let me know what you think.

FestiveCat: Hmm, I'm not particularly fond of using String::split as it relies heavily on the presence of the regex given. This is especially the case where I would need to split multiple times, where I need to check for the length multiple times in addition to creating many objects and eventually enumerating through all the non-separators/dividers to ensure that they are all numbers/digits.

I do agree with how it will remove the magic numbers though, but I think I will keep to this implementation for now? Since I did explain the magic numbers and they aren't too hard to understand

Thank you so much for your input though! I really appreciate it!

PR #120, opened 2022-01-23
[Zhou GuanZong] ip

Gnoznaug:
# Duke Pro
🔥 

> “Your mind is for having ideas, not holding them.” – David Allen

Some features include
- Adding tasks
- Deleting tasks
- Marking tasks

Example:
```
1. [X] Read book
2. [ ] Return book (by: Feb 05 2022)
```
Thing to add:
- [x] PR description
- [ ] PR review
- [ ] Increments

Link to my repo:
[Here](https://github.com/Gnoznaug/ip)

Main method:
```Java
}
```
PR update requirements:
~~1. a heading~~
~~2. a bullet list~~
~~3. a numbered list~~
~~4. a fenced code block (with syntax highlighting)~~
~~5. a task list~~
~~6. an emoji~~
~~7. a blockquote~~
~~8. a hyper link~~
~~9. inline code~~
~~10. some text formatting: bold, italic, stikethrough etc.~~


src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,61 @@
+package duke;
+
+import java.io.EOFException;
+
+import duke.command.Command;
+import duke.exception.DukeException;
+import duke.manager.Parser;
+import duke.manager.Storage;
+import duke.manager.TaskList;
```
iztanpy, line 9: nicely done on making sure not to use generic import *

src/main/java/duke/command/AddDeadlineCommand.java
```diff
@@ -0,0 +1,58 @@
...
+
+import duke.exception.DukeException;
+import duke.manager.Storage;
+import duke.manager.TaskList;
+import duke.task.Deadline;
+
+/**
+ * Represents a command that will add a Deadline Task to the TaskList upon execution..
+ */
+public class AddDeadlineCommand extends Command {
```
iztanpy, line 11: An alternative could be to have all the commands in one class, but I can understand that this might be neater too! :)

src/main/java/duke/command/MarkCommand.java
```diff
@@ -0,0 +1,62 @@
...
+            response += "Nice! I've marked this task as done:" + "\n";
+        } else {
+            taskList.markUndone(taskNo);
+            response += "OK, I've marked this task as not done yet:" + "\n";
+        }
+        response += taskList.getTask(taskNo).toString();
+        try {
+            storage.save(taskList);
+        } catch (Exception e) {
+            throw new DukeException(e.getMessage());
```
iztanpy, line 50: Nice! I think that this is more intuitive, I was previously typing out unique error messages for each error but I think this is a smarter solution!

iztanpy (commented):
I like the use of classes, and how this project utilised oop to the max! :))

PR #127, opened 2022-01-23
[Roh Yong Gi] iP

robinrojh:
Duke Chatbot
======

> "Huh?!" - Me, when my code is working.

**Duke chatbot** can serve as a nice time management tool for your life:
- Basic Todo list
- Deadline-based list with due dates
- Event-based list with event time
- _Sort_ your list based on date and content
- _Find_ your tasks using the name of the task

You can test out the application.
1. Use this [link](https://github.com/robinrojh/ip/releases/tag/A-Jar) to download!
2. Open the file using Java.
3. Add tasks using text! 
Some Commands are the following:
```
list
todo taskName
deadline taskName \by dueDate
event taskName \at atDate
```
Other features to be completed:
- [ ] Managing tasks
- [ ] GUI
- [ ] More testing for stability

There's a lot more coming😉!

-----
Very simple main method.
```Java
        Duke duke = new Duke(filePath);
        duke.run();
}
```


src/main/java/Duke.java
```diff
@@ -1,10 +1,64 @@
...
+import duke.exception.DukeException;
+import duke.parser.Parser;
+import duke.storage.Storage;
+import duke.task.TaskList;
+import duke.ui.Ui;
+
+import java.io.FileNotFoundException;
+import java.time.format.DateTimeFormatter;
+import java.time.format.DateTimeFormatterBuilder;
+import java.util.*;
```
Jiaaa-yang: The coding standard suggests to use single class imports instead of wildcard imports!

src/main/java/Duke.java
```diff
@@ -1,10 +1,64 @@
...
+
+    private Storage storage;
+    private TaskList tasks;
+    private Ui ui;
+
+    /**
+     * Loads the Duke chatbot with a given filePath to the saved file.
+     * If there is no file in the filePath, Storage will create a new file.
+     * @param filePath File's path.
+     */
```
Jiaaa-yang: The coding standard suggests to add a line break between description and parameter section. This might have been forgotten since it seems that there are line breaks for other comments.

src/main/java/duke/command/AddCommand.java
```diff
@@ -0,0 +1,52 @@
...
+     * @param command Object Another command object to be compared to.
+     * @return boolean Boolean value to show if this equals command.
+     */
+    @Override
+    public boolean equals(Object command) {
+        if (this.task.compareTo(((AddCommand) command).task) == 0) {
+            return true;
+        } else {
+            return false;
+        }
```
Jiaaa-yang, line 50:
```suggestion
        return this.task.compareTo(((AddCommand) command).task) == 0
```

Do you think the above suggestion would make this method more readable? I believe this is also linked to a particular antipattern which you can read up more on [here](https://ucsb-cs16.github.io/topics/antipatterns/)

src/main/java/duke/command/ListCommand.java
```diff
@@ -0,0 +1,42 @@
...
+
+    /**
+     * Checks the equality of two Command objects based on commandArray index 0.
+     *
+     * @param command Object Another command object to be compared to.
+     * @return boolean Boolean value to show if this equals command.
+     */
+    @Override
+    public boolean equals(Object command) {
+        return this.getCommandArray()[0].equals(((ListCommand) command).getCommandArray()[0]);
```
Jiaaa-yang, line 40: I think it is a little hard to read this line. You could consider making it more verbose using some variables!

src/main/java/duke/command/MarkCommand.java
```diff
@@ -0,0 +1,61 @@
...
+import duke.ui.Ui;
+
+/**
+ * Mark Command, in charge of marking a task as done/undone.
+ */
+public class MarkCommand extends Command {
+    public enum Mark {
+        MARK,
+        UNMARK;
+    }
```
Jiaaa-yang, line 15: I think it is an interesting idea to use an enum to differentiate between mark and unmarking commands! I personally had 2 classes (`MarkCommand.java` and `UnmarkCommand.java`) which resulted in a lot of repetition.

src/main/java/duke/parser/Parser.java
```diff
@@ -0,0 +1,147 @@
...
+        String firstArg = inputArray[0];
+        if (inputArray.length == 1) {
+            if (Command.CommandType.BYE.equals(firstArg)) {
+                return new ByeCommand(inputArray);
+            } else if (Command.CommandType.LIST.equals(firstArg)) {
+                return new ListCommand(inputArray);
+            } else {
+                throw new DukeMissingArgumentException(firstArg);
+            }
+        } else {
```
Jiaaa-yang, line 203: The condition under this loop seems to quite heavily nested. Do you think it would be better to abstract out the conditions to use a `switch` statement? 
Then maybe the logic relevant to the specialised checking for each type of task could also be moved to a separate class/done by the **Task** classes themselves?

Jiaaa-yang (commented):
Hey I think this PR is looking good! Other than a few minor violations of the given coding standard (which could be overlooked as you did adhere to them for the majority of your code), the code is neat and the logic is separated by relevant classes cleanly! I also liked that you have implemented features beyond the current requirements, like the sort command.

src/main/java/duke/command/ByeCommand.java
```diff
@@ -0,0 +1,41 @@
...
+import duke.task.TaskList;
+import duke.ui.Ui;
+
+/**
+ * Bye Command, in charge of bye command.
+ */
+public class ByeCommand extends Command {
+    /**
+     * Instantiates a ByeCommand with a commandArray.
+     * @param commandArray String[] String array of input command.
```
ryanwalterlee, line 13: Very minor coding standard violation here, there should be 1 line in between the description and the parameter section as stated here [link](https://se-education.org/guides/conventions/java/intermediate.html)

src/main/java/duke/command/Command.java
```diff
@@ -0,0 +1,71 @@
...
+         *
+         * @param input String Input String to be compared
+         * @return boolean Boolean value to show if this equals command.
+         */
+        public boolean equals(String input) {
+            if (input.equalsIgnoreCase(this.name())) {
+                return true;
+            } else {
+                return false;
+            }
```
ryanwalterlee, line 40: To improve readability of code, I might write it this way
```
if (input.equalsIgnoreCase(this.name())) return true;
return false
```
This shortens the code to half the number of lines which can make it less daunting to read. If you like this version instead feel free to try it.

src/main/java/duke/parser/Parser.java
```diff
@@ -0,0 +1,160 @@
...
+                    return new SortCommand(TaskList.SortType.CONTENT, inputArray);
+                } else if (TaskList.SortType.DATE.equals(secondArg)) {
+                    return new SortCommand(TaskList.SortType.DATE, inputArray);
+                } else {
+                    throw new DukeInvalidArgumentException("sort type");
+                }
+            } else {
+                return new InvalidCommand();
+            }
+        }
```
ryanwalterlee: Personally I used Java String's startsWith method to determine the first word of the sentence. I get what you are trying to do here but it is quite daunting to read many nested ifs and elses, do you agree?

src/main/java/duke/task/Task.java
```diff
@@ -0,0 +1,118 @@
...
+
+import java.time.LocalDateTime;
+
+import duke.ui.Ui;
+
+
+public class Task implements Comparable<Task> {
+    protected boolean isDone;
+    protected String content;
+    protected LocalDateTime date;
```
ryanwalterlee, line 12: Should these be protected? I think we can get away with private as we would not want other classes in this package to be able to access this right?

ryanwalterlee (commented):
Good modularisation by splitting actions into many classes.

1 thing you can improve on would be to simplify the parts of the code that has very long ifs and elses as I see it somewhat often.

PR #132, opened 2022-01-24
[Nguyen Hong Anh] IP

honganhcs:
# AnnBot
> “Life is too complicated not to be orderly.” – Martha Stewart 

AnnBot frees your mind of having to remember things you need to do. It is:
- text-based
- easy-to-learn
- ~~FAST~~ SUPER FAST to use 😉 
All you need to do is:
1. download it from [here](https://github.com/honganhcs/ip/releases)
2. double-click it
3. add your tasks 
4. let it manage your tasks for you 😄 
And it is **FREE**!

Features:
- [x] Managing tasks, deadlines and events
- [x] Searching for tasks by keywords
- [ ] Reminders (coming soon)

Here's the ` main ` method:
``` java
        new Ann().run(args);
}
```


src/main/java/ann/commands/AddCommand.java
```diff
@@ -0,0 +1,60 @@
...
+        switch (taskType) {
+            case TODO:
+                task = new Task(taskComponents[0]);
+                break;
+            case DEADLINE:
+                task = new Deadline(taskComponents[0], taskComponents[1]);
+                break;
+            case EVENT:
+                task = new Event(taskComponents[0], taskComponents[1]);
+        }
```
Jiaaa-yang: The [module coding standards](https://se-education.org/guides/conventions/java/intermediate.html#layout) suggests to not indent `case` statements!

src/main/java/ann/parser/Parser.java
```diff
@@ -0,0 +1,201 @@
...
+            return handleMark(input.substring(4));
+        } else if(input.length() < 6) {
+            return new IncorrectCommand("Oops! Please enter a valid command!");
+        } else if(input.substring(0, 6).toLowerCase().equals("unmark")) {
+            return handleUnmark(input.substring(6));
+        } else if(input.substring(0, 6).toLowerCase().equals("delete")) {
+            return handleDelete(input.substring(6));
+        } else {
+            return new IncorrectCommand("Oops! Please enter a valid command!");
+        }
```
Jiaaa-yang:
```java
if (input.length() < 3) {
    // Check commands where 0 <= length < 3
} else if (input.length() < 4) {
    // Check commands where 3 <= length < 4
}
.
.
.
```

Do you think this portion will be more readable if it was changed to the above format? Personally I believe this block of code is not very straightforward as it seems some clauses depend on a certain length of input. Using the length as if-else conditions may help reinforce this idea.

src/main/java/ann/parser/Parser.java
```diff
@@ -0,0 +1,201 @@
...
+    /**
+     * Parses the contents of an 'add' command and returns the appropriate Command object.
+     *
+     * @param input a String which represents the contents of the 'add' command.
+     * @return an appropriate Command object.
+     */
+    private static Command handleAdd(String input) {
+        if(input.isBlank()) {
+            return new IncorrectCommand("Oops! Please specify the task you wish to add!");
+        } else if(input.charAt(0) != ' ') {
```
Jiaaa-yang: The coding standard suggests a space after an `if` or `else if`!  It seems that this is followed in some places but not elsewhere, so maybe you could have overlooked it.

src/main/java/ann/storage/Storage.java
```diff
@@ -0,0 +1,156 @@
...
+    }
+
+    /**
+     * Represents an Exception thrown when a given filename is not valid.
+     */
+    public static class InvalidStorageFilePathException extends IllegalArgumentException {
+        public InvalidStorageFilePathException(String message) {
+            super(message);
+        }
+    }
```
Jiaaa-yang, line 146: I like that this particular exception extends from `IllegalArgumentException`! Since it also represents some form of invalid argument, and this exception also allows clearer error handling with regards to the Storage class for your AnnBot!

src/main/java/ann/commands/FindCommand.java
```diff
@@ -0,0 +1,14 @@
+package ann.commands;
+
+public class FindCommand extends Command{
+    private String findKeyWords;
```
Jiaaa-yang:
```suggestion
    private String keywords;
```

Do you think it would be more appropriate if this variable was renamed as above? Since 'find' is a verb which can suggest a method style of naming. This could be confusing such as in the `executeCommand` below where you pass in `this.findKeyWords` as an argument, which may seem like a method call.

Jiaaa-yang (commented):
Hey I think this PR is looking good! Your implementations are neat with logic properly separated into relevant classes and packages. I also like how some classes are short and do just what their name implies they should! Except for a few minor style errors, and some `if` statements with a lot of `else-if`, I think your code is highly readable!

src/main/java/ann/commands/AddCommand.java
honganhcs: Hi! Thanks a lot for your comments, you pointed out some issues that I overlooked and I will definitely fix them 😄

src/main/java/ann/parser/Parser.java
honganhcs: Actually for this, it's not just the length I have to check, but also whether it matches the command pattern, e.g. it can be shorter than 4 characters but if it is "bye" then I have to accept the command. I have to check the length in between as substring() would throw an index out of bounds exception otherwise. I'm still thinking about how this could be improved 😄

src/main/java/ann/commands/FindCommand.java
honganhcs: Thanks for pointing this out! I'll fix it 😄

PR #154, opened 2022-01-24
[Tan Weiu Cheng] iP

DavidTan0527:
# Nikki

> Rule 35 of JS: if it exists, there is a JavaScript framework with that name

Nikki is a *terminal-based application* for you to manage your tasks ~like a pro manager (jkjk)~

## Installation Guide
1. Download it from [here](https://github.com/DavidTan0527/ip/releases/tag/A-Jar).
2. Open your favorite terminal application.
3. Run `java -jar path/to/file.jar`
4. Enjoy! 🤗 

## Features
- Create, Read, Delete tasks
- Mark tasks as done or not
- Read and write to file for saving tasks in permanent storage
- 3 types of tasks: Deadline, Event, Todo
- Syntax checking and parsing for commands

## ToDos 🗒️ 
- [ ] Add command subtypes to `Command` for better tokenization
- [ ] Write command list for user guide

---

## Questions
*Note: this section is for nerds*

### Command tokenization
Currently, I'm using `regex` to parse my commands. For example, below is the regex for parsing the command `deadline`

```regex
deadline (.*) /(by) ((?:[1-9]|[12][0-9]|3[01])/(?:[1-9]|1[0-2])/[0-9]{4})
```

which is used as follows[^parser]

```java
Pattern p = Pattern.compile(/* regex above */);
Matcher m = p.matcher(cmd);

if (m.matches()) {
    if (m.groupCount() > 0) {
        String arg = m.group(1);
        HashMap<String, String> kwargs = new HashMap<>();

        // kwargs start at 2,
        // The last one is groupCount() - 1, its value being groupCount()
        for (int i = 2; i < m.groupCount(); i += 2) {
            kwargs.put(m.group(i), m.group(i + 1));
        }

        return new Command(cmdName, arg, kwargs);
    }
    else {
        return new Command(cmdName);
    }
} else {
    // Syntax error
    ...
}
```

The motivation of such design is for scaling for commands such that there can be more keyword arguments and it can be in arbitrary order.

**Question**: Is there a better way to achieve that result?

[^parser]: found in [`src/main/java/nikki/command/CommandParser.java`](https://github.com/DavidTan0527/ip/blob/master/src/main/java/nikki/command/CommandParser.java)

### File I/O handling
Since I'm reading from and writing to file, I would need to check its existence. My current solution is to create a file if it doesn't exist upon reading.

The caveat is when the user has ***no** write permission* in the directory of execution, then the file is not created, which causes issues when writing to the file. To mitigate this, I have also put the same file check + creation routine in the method that writes to the file.

However, it feels redundant to repeat the same procedure as it would've succeeded the first time and wouldn't need a second time. Alternatively, the second would've failed for the same reason as the first time.

**Question**: I'm not sure if my current approach[^file] is fine as it is or if there're any better solutions.

[^file]: found in [`src/main/java/nikki/Storage.java`](https://github.com/DavidTan0527/ip/blob/master/src/main/java/nikki/Storage.java)


src/main/java/nikki/task/Task.java
```diff
@@ -0,0 +1,204 @@
+package nikki.task;
+
+import nikki.NikkiException;
+
+public class Task {
+    /** Mark in checkbox of completed Task */
+    private static final char completedMark = 'X';
```
goel-a: Could use uppercase variable names to store constant

src/main/java/nikki/task/Task.java
```diff
@@ -0,0 +1,204 @@
...
+
+    /** Mark in checkbox of uncompleted Task */
+    private static final char incompleteMark = ' ';
```
goel-a: Could use uppercase variable names to store constant

src/main/java/nikki/task/Task.java
```diff
@@ -0,0 +1,204 @@
...
+                this.isDone() ? Task.completedMark : Task.incompleteMark,
+                this.name);
+    }
+
+    /**
+     * Returns Task info in a standard format for saving in file.
+     *
+     * @return formatted string for saving Task
+     */
+    public String fileSaveFormat() {
```
goel-a: Could use a verb for the name of the method

src/main/java/nikki/NikkiException.java
```diff
@@ -0,0 +1,7 @@
+package nikki;
+
+public class NikkiException extends Exception {
+    public NikkiException(String errorMessage) {
```
goel-a, line 9: Could consider adding a header comment to describe what kind of exceptions are being handled by this class/method

src/main/java/nikki/task/TaskList.java
```diff
@@ -0,0 +1,185 @@
...
+    public int size() {
+        return this.tasks.size();
+    }
+
+    /**
+     * Returns a string, where each line is a Task formatted for saving to file
+     *
+     * @return formatted string of Tasks for saving to file
+     */
+    public String tasksFileSaveFormat() {
```
goel-a: Could consider naming the method using a verb

goel-a (commented):
Looks good! 
Some cosmetic changes recommended

src/main/java/nikki/task/Task.java
DavidTan0527: Thanks for pointing this out!

src/main/java/nikki/task/Task.java
DavidTan0527: Will do, thanks!

PR #159, opened 2022-01-24
[Yusuf Bin Musa] iP

yusufaine:
# DukePro
> "Your mind is for having ideas, not holding them." – David Allen ([Source](https://dansilvestre.com/productivity-quotes))

DukePro frees your mind of having to remember things you need to do. 

It's:
* 💻  text-based
* 🧑‍🎓  easy to learn
* 🚀  ~FAST~ _SUPER_ FAST to use

All you need to do is,
1. download the latest version [here](https://github.com/yusufaine/ip/releases),
2. ~run it in your terminal~ double-click it 😏 
3. add your tasks

And the best part -- it's **FREE**!

## Features:
- [x] ✅  Managing tasks
- [x] ⏰  Managing deadlines
- [x] 📅  Events or reminders

### Feature pipeline
<details>
<summary> Click here if you're interested to see what's in the pipeline 😉  </summary>

- [ ] Show Tasks based on their types (todo/deadlines/reminder)
- [ ] Sort events and deadlines by their stated time to occur/due
- [ ] Mark/unmark/remove several entries at once
- [ ] Add command aliasing (lookup table)
</details>

---
```java
public class Main {
    }
}
```


src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,104 @@
...
+public class Parser {
+
+    /**
+     * Parses `String` to `Integer`.
+     *
+     * @param strToIntValue the `String` that needs to be parsed to an `Integer` value
+     * @return the `Integer` value of the `String`.
+     * @throws DukeException if the parameter could not be parsed to an `Integer`.
+     */
+    public static int parseInt(String strToIntValue) throws DukeException {
```
DavidTan0527, line 31: Why did you create a method instead of using the built-in Java method [`Integer.parseInt(String s)`](https://docs.oracle.com/javase/10/docs/api/java/lang/Integer.html#parseInt(java.lang.String))?

src/main/java/duke/command/ListCommand.java
```diff
@@ -0,0 +1,46 @@
...
+            StringBuilder sb = new StringBuilder(ui.STR_PADDING + "Here are the tasks in your list: \n");
+            int[] taskCount = {1};
+
+            tasks.getTasks().forEach( task -> {
+                sb.append(String.format(ui.STR_PADDING + "  %d. " + task, taskCount[0]++));
+
+                if (taskCount[0] != tasks.size() + 1) {
+                    sb.append("\n");
+                }
+            });
```
DavidTan0527: Perhaps, we can reinforce the concept of abstraction by putting this as a method in `TaskList` class

src/main/java/duke/command/FindCommand.java
```diff
@@ -0,0 +1,47 @@
...
+    }
+
+    @Override
+    public void execute(TaskList tasks, Ui ui, Storage storage) {
+        if (tasks.isEmpty()) {
+            ui.printWithDivider("Your list is empty!");
+        } else {
+            List<Task> filteredTask = tasks.getTasks().stream()
+                    .filter(task -> task.getTaskName().contains(findString))
+                    .collect(Collectors.toList());
```
DavidTan0527, line 43: It might be better to abstract this method within `TaskList`, which might also work better with the code below.

src/main/java/duke/command/FindCommand.java
```diff
@@ -0,0 +1,47 @@
...
+            if (filteredTask.isEmpty()) {
+                ui.printWithDivider(String.format("I could not find any tasks that contains \"%s\"", findString));
+            } else {
+                StringBuilder sb = new StringBuilder(ui.STR_PADDING + "You've searched for: \"" + findString + "\"\n");
+                sb.append(ui.STR_PADDING + "Here are the matching tasks in your list:");
+
+                int[] taskCount = {1};
+                filteredTask.forEach(task -> {
+                    sb.append(String.format("\n" + ui.STR_PADDING + "  %d. " + task, taskCount[0]++));
+                });
```
DavidTan0527: Perhaps, we can reinforce the concept of abstraction by putting this as a method in `TaskList` class

src/main/java/duke/command/FindCommand.java
```diff
@@ -0,0 +1,47 @@
...
+                    .filter(task -> task.getTaskName().contains(findString))
+                    .collect(Collectors.toList());
+
+            if (filteredTask.isEmpty()) {
+                ui.printWithDivider(String.format("I could not find any tasks that contains \"%s\"", findString));
+            } else {
+                StringBuilder sb = new StringBuilder(ui.STR_PADDING + "You've searched for: \"" + findString + "\"\n");
+                sb.append(ui.STR_PADDING + "Here are the matching tasks in your list:");
+
+                int[] taskCount = {1};
```
DavidTan0527: `int` would suffice.
```suggestion
                int taskCount = 1;
```

src/main/java/duke/command/FindCommand.java
```diff
@@ -0,0 +1,47 @@
...
+                filteredTask.forEach(task -> {
+                    sb.append(String.format("\n" + ui.STR_PADDING + "  %d. " + task, taskCount[0]++));
```
DavidTan0527:
```suggestion
                    sb.append(String.format("\n" + ui.STR_PADDING + "  %d. " + task, taskCount++));
```

DavidTan0527 (commented):
Mostly great code design, just some suggestions that might help simplify your code. Please take a look. 😄

src/main/java/duke/Parser.java
yusufaine: Regarding this point my rationale was that I want the errors to be thrown as and where they happen when I was implementing the parts that needed `parseInt()` which was `DeleteCommand`, and `ToggleCommand`. 

In those commands I've also thrown in out-of-bounds errors, in those 2 files I've also done some "error-checking" as UI was available to me there and I was not keen on passing UI into the Parser (since parser should only really be processing strings)

src/main/java/duke/command/ListCommand.java
yusufaine: Oh, I like this idea! Not sure why I didn't think of moving it into `TaskList`, but it makes more sense that way -- thank you!

src/main/java/duke/command/FindCommand.java
yusufaine: Similar to `ListCommand`, I think this is really helpful 👍

src/main/java/duke/command/FindCommand.java
yusufaine: I believe this would throw an error as I used a lambda to check for the filtered tasks, if this was a (enhanced) for loop, it would work, but it's not possible to change local variables in lambda as they are `final or effectively final` -- see screenshot attached.

image

src/main/java/duke/Parser.java
DavidTan0527: Ahh, I see, it makes more sense now. Thanks for the explanation!

src/main/java/duke/command/FindCommand.java
DavidTan0527: Perhaps a traditional `for` loop would solve this issue 🤔

yusufaine: yep, I could've done an enhanced for-loop (or a normal one) and iterate over, but part of me really wanted to use a lambda here 😆  I'm not quite sure what the consensus is about handling it that way, but to play it safe I'll probably change it to a for-loop -- thanks for the review!

PR #161, opened 2022-01-24
[LeeEnEn] iP

LeeEnEn:
Duke Bot frees your mind of having to remember things you need to do. It is,
- text-based
- easy to learn
- ~FAST~ _SUPER_ FAST to use

All you need to do is,
1. download it from [here](https://github.com/nus-cs2103-AY2122S2/ip)
2. double-click it
3. add your tasks
4. let it manage your tasks for you :smiley:

And it is **FREE!**

Features
- [x] Managing tasks
- [ ] Managing deadline (soon)
- [ ] Reminders (soon)

``` ruby
public class Main {
    }
}
```


src/main/java/duke/parser/Parser.java
```diff
@@ -0,0 +1,194 @@
...
+     * Handles user command and give an appropriate reply.
+     * @param userInput commands specified by the user.
+     */
+    public void parse(String userInput) {
+        String[] input = userInput.split(" ", 2);
+
+        Type type = getEnumType(input[0]);
+        TaskList taskList = new TaskList();
+
+        try {
```
channne, line 33: I really like how readable this block is, as well as how you printed different types of error messages for different types of errors :smiley:

src/main/java/duke/parser/Parser.java
```diff
@@ -0,0 +1,194 @@
...
+
+            if (task == null || userInput[1].strip().isEmpty()) {
+                throw new DukeException("Error in task");
+            }
+            return task;
+
+        } catch (DukeException e) {
+            if (isCommandRecognized(userInput[0])) {
+                ui.echo("☹ OOPS!!! The description of a " + userInput[0] + " cannot be empty.");
+            } else if (e.getMessage().equals("Error in task") || e.getMessage().equals("Error")) {
```
channne: Perhaps it would be better to create different classes of exceptions instead of recognizing the different types of exceptions by their error messages?

src/main/java/duke/task/Task.java
```diff
@@ -0,0 +1,88 @@
...
+
+import duke.ui.Ui;
+
+import java.time.LocalDate;
+
+/**
+ * Handles the creation of tasks.
+ */
+public class Task {
+    private boolean marked;
```
channne: Consider renaming the booleans in this class?

src/main/java/duke/parser/Parser.java
```diff
@@ -0,0 +1,194 @@
...
+        }
+        return task;
+    }
+
+    /**
+     * Returns a boolean depending on Duke understanding the given command.
+     * @param input User command.
+     * @return True if Duke understands the command, false otherwise.
+     */
+    private boolean isCommandRecognized(String input) {
```
channne: Perhaps this method could be better named, since it detects four specific types of commands rather than all commands

src/main/java/duke/storage/Storage.java
```diff
@@ -0,0 +1,94 @@
...
+                }
+
+                if (temp[1].equals("X")) {
+                    task.markTask(true, false);
+                }
+                list.add(task);
+            }
+            scanner.close();
+        } catch (FileNotFoundException e) {
+            System.out.println("OOPS!!! Duke.Main.Duke is unable to locate your file!");
```
channne: Perhaps it would be better to print error messages through the UI as well?

channne (commented):
Overall, I think your code is very readable and I could hardly find any coding standard violations :+1: One thing I noticed however is that many classes communicate with the UI! Perhaps it might be beneficial to restructure such that fewer classes communicate with it :smile:

src/main/java/duke/parser/Parser.java
LeeEnEn: Thank you! 😄

src/main/java/duke/parser/Parser.java
LeeEnEn: That makes sense, I will do the necessary changes!

src/main/java/duke/task/Task.java
LeeEnEn: Good catch! Definitely missed this during my checks!

src/main/java/duke/parser/Parser.java
LeeEnEn: You are right, thank you for pointing that out! 😄

src/main/java/duke/storage/Storage.java
LeeEnEn: This makes sense too! Thank you for the feedback! 👍

src/main/java/duke/parser/Parser.java
```diff
@@ -0,0 +1,194 @@
...
+        String temp = input.toLowerCase();
+        return temp.equals("todo") || temp.equals("event") || temp.equals("deadline") || temp.equals("delete");
+    }
+
+    /**
+     * Returns a boolean that signifies the user's wish to end the program.
+     * @return A boolean.
+     */
+    public boolean isExit() {
+        return this.isExit;
```
brandonrhan: You never use "this." in previous code, maybe you could delete this one as well

src/main/java/duke/task/Deadline.java
```diff
@@ -0,0 +1,32 @@
...
+
+import java.time.LocalDate;
+import java.time.format.DateTimeFormatter;
+
+/**
+ * Represents a Deadline task.
+ */
+public class Deadline extends Task {
+
+    private LocalDate by;
```
brandonrhan: Maybe you should rename it using a more detailed name

src/main/java/duke/task/Deadline.java
```diff
@@ -0,0 +1,32 @@
...
+    }
+
+    /**
+     * Prints out the details of the Deadline task.
+     * @return A string with details of the Deadline task.
+     */
+    @Override
+    public String toString() {
+        return "[D]" + super.toString()
+                + " (by: " + by.format(DateTimeFormatter.ofPattern("MMM d yyyy")) + ")";
```
brandonrhan: The pattern could be saved as a static variable to avoid duplicate code

src/main/java/duke/task/Event.java
```diff
@@ -0,0 +1,32 @@
...
+
+import java.time.LocalDate;
+import java.time.format.DateTimeFormatter;
+
+/**
+ * Represents an Event task.
+ */
+public class Event extends Task {
+
+    private LocalDate at;
```
brandonrhan: at is also not very clear, I think could change to another name

src/main/java/duke/task/Task.java
```diff
@@ -0,0 +1,88 @@
...
+        this.charId = charId;
+        this.date = date;
+    }
+
+    /**
+     * Marks the task as completed or incomplete.
+     * @param mark A boolean to mark or unmark a task.
+     * @param show A boolean to print Duke's reply to the user.
+     */
+    public void markTask(boolean mark, boolean show) {
```
brandonrhan: show should also be renamed

brandonrhan (commented):
I really like your code!! Your naming of variables is so good and your code is extremely clear to read. I have learned a lot from you!!

PR #168, opened 2022-01-24
[Tan Yu Meng] iP

yumengtan:
# **Lucifer** 😈

> Just think of me as your hot, sexy, irresistible guardian devil. - Lucifer, probably

Are you down bad like [these people](https://www.youtube.com/watch?v=fbVRnlR_6BM&ab_channel=BuzzFeedCeleb)?
Fret not! Your favourite devil is now here to serve you miscreants!
Fantasize it as your virtual Tom Ellis that assists you with your desires. It's:

- text-based
- easy to learn

All you need to do is,

1. download it from [here](https://github.com/yumengtan/ip/archive/refs/tags/A-Jar.zip).
2. double-click it.
3. add your desires.
4. let it manage your desires for you 👅

And it is **FREE!**

Features: 
- [x] Managing desires
- [x] Managing deadlines
- [ ] Reminders (coming soon)
- [ ] And much more (coming soon)

You can use it to practice Java programming too. Here's the main method:
```Java
        new Lucifer().run();
    }
```


src/main/java/Commands/AddCommands.java
```diff
@@ -0,0 +1,77 @@
...
+import TaskList.TaskList;
+
+import java.time.LocalDateTime;
+import java.util.ArrayList;
+
+/**
+ * Class AddCommmands which inherits from ParseCommands.
+ * @author Yu Meng
+ */
+public class AddCommands extends ParseCommands{
```
Geralddtan: Perhaps there should be a " " separator before the {. I have also noticed the same issue in some other places, notably in the AddCommandsTest.java file!

src/main/java/Commands/AddCommands.java
```diff
@@ -0,0 +1,77 @@
...
+        this.list = task.getTaskList();
+    }
+
+    /**
+     * Adds the task to the list if it is of type Event.
+     *
+     * @param desc the description of task
+     * @param old the user input format of the event date
+     */
+    public void addEvent(String desc, LocalDateTime old) {
```
Geralddtan: Perhaps a more intuitive variable name here instead of "old"? This might give increased clarity to the reader!

src/main/java/Commands/AddCommands.java
```diff
@@ -0,0 +1,77 @@
...
+     * @param by the date and time of the task deadline
+     */
+    public void addDeadline(String description, LocalDateTime by) {
+        System.out.println(UNDERSCORE);
+        System.out.println("Got it. I have added this to your desires:");
+        Deadline curr = new Deadline(description, by);
+        list.add(curr);
+        System.out.println(curr);
+        System.out.println("Currently you have " + list.size() + " things yet to be desired");
+        System.out.println(UNDERSCORE);
```
Geralddtan: Perhaps an abstraction method which performs the task of printing the UNDERSCORE lines might be a good addition! I've noticed several cases of this being repeated and abstracting this out to a separate method/class could be better!

src/main/java/Commands/ParseCommands.java
```diff
@@ -0,0 +1,129 @@
...
+     * @param task the class that stores the list of tasks
+     */
+    public ParseCommands(TaskList task) {
+        this.task = task;
+    }
+
+    /**
+     * Parses the command from input.
+     */
+    public void parseCommand(String command) {
```
Geralddtan: This method seems a little lengthy and is deeply nested, making it a little difficult to understand the proper flow. Perhaps a more concise method/abstraction methods could be employed!

src/main/java/Lucifer.java
```diff
@@ -0,0 +1,69 @@
...
+
+import java.io.IOException;
+
+import java.util.Scanner;
+
+  /**
+ * Main class which runs the Lucifer chatbot
+ * @author Yu Meng
+ */
+ public class Lucifer {
```
Geralddtan: I really like the idea of using Lucifer :) Gives your task handler a great persona!

Geralddtan (commented):
Overall, I think your code structure is well organised! However, some methods seem pretty length and deeply nested, making it hard to decipher the logical flow. You also have followed the coding style practices quite closely with just a few issues pointed out such as having more informative variable names! Overall, LGTM, just a few nits to fix! Love the idea of using Lucifer!

PR #172, opened 2022-01-25
[Tonishka Singh] iP

tonishka:
### YodaBot 👾 👾 

> "Do or do not. There is no try." - Yoda ([More about Yoda](https://www.starwars.com/databank/yoda))

If Star Wars fan you are, use this bot you must.  It is: 
- text-based
- easy to learn
- ~~fast~~ SUPER FAST to use
- in the **wise voice** of Yoda

YodaBot can: 
- [X] Manage and store your tasks
- [X] Categorise tasks based on type and status
- [X] Search for target tasks
- [ ] Send reminders (Soon 😃) 

`java -jar Yoda.jar`

```java
public static void main() {
    System.out.println("Wishes you good day, Yoda does.");
}
```


src/main/java/yoda/QuestList.java
```diff
@@ -0,0 +1,62 @@
...
+
+/**
+ * Represents a list of all existing quests.
+ * @author Tonishka Singh
+ */
+
+import java.util.ArrayList;
+
+public class QuestList {
+    private ArrayList<Quest> questList;
```
muraddurrani, line 10: Should the name of this list be quests? I think the coding standards suggests that lists containing several items be named as a plural object.

src/main/java/yoda/Deadline.java
```diff
@@ -0,0 +1,23 @@
+package yoda;
+
+/**
+ * Represents a deadline bound quest.
+ * @author Tonishka Singh
+ */
+
```
muraddurrani: I think there should not be a blank line between a JavaDoc comment and the method/class that it precedes. I noticed this in a few other places too.

src/main/java/yoda/Parser.java
```diff
@@ -0,0 +1,115 @@
...
+ * @author Tonishka Singh
+ */
+
+public class Parser {
+    public Parser() {}
+
+    /**
+     * Tokenizes and cleans the input by removing extraneous spaces.
+     * @param input Input string from user to clean.
+     * @return Array of cleaned String arguments.
```
muraddurrani, line 13: I think there should be an empty line between the description and the parameters section. I noticed this in some other comments too.

src/main/java/yoda/Parser.java
```diff
@@ -0,0 +1,115 @@
+package yoda;
+
+/**
+ * Parser object to make sense of user command.
+ * @author Tonishka Singh
+ */
+
+public class Parser {
+    public Parser() {}
```
muraddurrani: Methods are supposed to be written with Egyptian brackets. Since this constructor is essentially default constructor, I suggest that you just delete it to avoid the trouble.

src/main/java/yoda/Quest.java
```diff
@@ -0,0 +1,93 @@
...
+        this.deadline = LocalDate.parse(date);
+        this.status = 0;
+        this.type = "D";
+    }
+
+    public void completeQuest() {
+        this.status = 1;
+    }
+
+    public void incompleteQuest() {
```
muraddurrani, line 53: I think method names should be phrased as verbs. Perhaps completeQuest and incompleteQuest could be changed to something like markasComplete and markasIncomplete?

src/main/java/yoda/Ui.java
```diff
@@ -0,0 +1,78 @@
+package yoda;
+
+/**
+ * Encapsulates the user interface for the chatbot.
+ * @author Tonishka Singh
+ */
+public class Ui {
+    public Ui() {}
```
muraddurrani: This is similar to the case with the constructor for Parser. I think you can just delete it to simplify things and avoid violations.

src/main/java/yoda/Yoda.java
```diff
@@ -0,0 +1,131 @@
...
+        }
+    }
+
+    private void log() {
+        System.out.println("To defeat the dark lord, " +
+                "following quests must you finish, Jedi:");
+        System.out.print(questList.toString());
+    }
+
+    private void todo(String desc) {
```
muraddurrani: Perhaps this could be changed to createTodo or makeTodo in order to make sure its phrased as a verb. The same applies for the following methods.

src/main/java/yoda/QuestList.java
```diff
@@ -0,0 +1,62 @@
...
+        questList.remove(questID);
+        return q;
+    }
+
+    public int numQuests() {
+        return questList.size();
+    }
+
+    public ArrayList<Quest> find(String keyword) {
+        ArrayList<Quest> result = new ArrayList<>();
```
muraddurrani, line 39: Perhaps the name could be changed to something like results or matches? List objects should be written as plural.

src/main/java/yoda/Storage.java
```diff
@@ -0,0 +1,112 @@
...
+
+    public Storage() {
+        this.dir = System.getProperty("user.dir");
+        this.filePath = dir + File.separator + "YodaData" + File.separator + "data.txt";
+        this.data = new File(this.filePath);
+    }
+
+    public ArrayList<Quest> load() {
+        Path path = Paths.get(dir, "YodaData");
+        boolean directoryExists = Files.exists(path);
```
muraddurrani, line 34: The coding standard suggests that booleans be named to sound like booleans, so perhaps something like doesDirectoryExist?

src/main/java/yoda/Ui.java
```diff
@@ -0,0 +1,78 @@
...
+
+    /**
+     * Prints the welcome message when user launches the application.
+     */
+    public void welcome() {
```
muraddurrani: Maybe call this method printWelcome or displayWelcome or something along those lines to phrase it as a verb?

muraddurrani (commented):
I love how you have committed to the Yoda angle and how you've changed tasks to quests! Adds a lot of personality to the interaction.
Generally, I found the code fairly intuitive to read. I think the main things to focus on are slightly altering the JavaDocs and renaming some of the methods and field names to follow standards.

src/main/java/Yoda.java
```diff
@@ -1,26 +1,37 @@
 import java.util.Scanner;
 
 public class Yoda {
+    private enum YodaCommands {
```
kavxya: I like how you made use of the enums for your bot.

src/main/java/Yoda.java
```diff
@@ -2,7 +2,7 @@
 
 public class Yoda {
     public static void main(String[] args) {
-
+        Jedi jedi = new Jedi();
```
kavxya: May I know what the Jedi object is for?

src/main/java/yoda/Ui.java
```diff
@@ -43,6 +48,9 @@ private static void type(String text) {
         }
     }
 
+    /**
+     * Prints out the help message for user.
+     */
     public void help() {
```
kavxya: Would the user benefit from receiving more statements in the help command?

PR #176, opened 2022-01-25
[Gareth Koh] iP

garethkoh:
- text-based
- easy to learn
All you need to do is,

2. double-click it.
3. add your tasks.
4. let it manage your tasks for you 💯 

its **FREE**

tasklist:
- [x] done
- [ ] not done

here is the main method:

```java
public class Duke {

        // load in the TaskList
        TaskList taskList = new TaskList(new Storage());
        Parser parser = new Parser(taskList);
        Ui.initialize(parser);
        Storage.save(taskList);

    }
```


src/main/java/Duke.java
```diff
@@ -1,10 +1,223 @@
...
+        }
+    }
+
     public static void main(String[] args) {
+
+        // Initialise global objects
+        Scanner sc = new Scanner(System.in);
+        ArrayList<Task> toDoList = new ArrayList<>();
+        try {
+
```
Korochii: Might want to delete this line (line 32) since it feels abit unneccesary

src/main/java/Duke.java
```diff
@@ -1,10 +1,223 @@
...
+            } else if (split[0].equals("help")) {
+                System.out.println("try the following commands:\n"
+                        + "   list (list out all tasks in your todo list)\n"
+                        + "   todo <task> (add a basic task to your todo list)\n"
+                        + "   deadline <task> /by <date> (add a task with a deadline to your todo list)\n"
+                        + "   event <task> /at <date> (add a new event with its corresponding date to your todo list)\n"
+                        + "   mark <task index> (mark that specific task as completed)\n"
+                        + "   unmark <task index> (remove mark from specific task)\n"
+                        + "   delete <task index> (remove task from todo list)\n"
+                        + "   bye (close application)");
```
Korochii: Interesting addition to Duke that was not specified in the basic requirements (and I did not think of implementing this too!)
Good job 😃

src/main/java/Duke.java
```diff
@@ -1,10 +1,223 @@
...
+                    } catch (IndexOutOfBoundsException e) {
+                        System.out.println(("the index you have entered does not exist!"));
+                    }
+                } catch (IndexOutOfBoundsException e) {
+                    System.out.println("unmark description cannot be empty!");
+                } catch (NumberFormatException e) {
+                    System.out.println("unmark description must be an integer!");
+                }
+
+
```
Korochii: Might want to delete these empty lines too so that the formatting is abit more consistent

src/main/java/Duke.java
```diff
@@ -1,10 +1,223 @@
...
+                        System.out.printf("great job! I've marked this task as done: %s\n", t.getName());
+                    } catch (IndexOutOfBoundsException e) {
+                        System.out.println(("the index you have entered does not exist!"));
+                    }
+                } catch (IndexOutOfBoundsException e) {
+                    System.out.println("mark description cannot be empty!");
+                } catch (NumberFormatException e) {
+                    System.out.println("mark description must be an integer!");
+                }
+
```
Korochii: Might want to delete this empty line too so that the formatting is abit more consistent

src/main/java/Duke.java
```diff
@@ -1,10 +1,223 @@
...
+                }
+            } else if (split[0].equals("deadline")) {
+                try {
+                    // Check if description is empty
+                    String check = split[1];
+                    try {
+                        // Take the substring of user input after deadline
+                        String item = userInput.substring(9);
+                        // Divide the substring into task name and deadline
+                        String[] divide = item.split("/");
```
Korochii: In this case, divide should be in plural form since it is an array of strings

src/main/java/Duke.java
```diff
@@ -1,10 +1,223 @@
...
+
+            } else if (split[0].equals("event")) {
+                try {
+                    // Check if description is empty
+                    String check = split[1];
+                    try {
+                        // Take the substring of user input after event
+                        String item = userInput.substring(6);
+                        // Divide the substring into task name and date
+                        String[] divide = item.split("/");
```
Korochii: In this case, divide should be in plural form since it is an array of strings

Korochii (commented):
Overall, most of the coding standards are being adhered to (which is why I had to nitpick a few minor issues :sweat_smile:)

Good job and keep up the good work!

src/main/java/Event.java
```diff
@@ -0,0 +1,30 @@
...
+
+    @Override
+    public String toString() {
+        if (super.getStatus() == 1) {
+            return "[E][X] " + super.getName() + " (at: "
+                    + time.format(DateTimeFormatter.ofLocalizedDate(FormatStyle.FULL)) + ")\n";
+        } else {
+            return "[E][ ] " + super.getName() + " (at: "
+                    + time.format(DateTimeFormatter.ofLocalizedDate(FormatStyle.FULL)) + ")\n";
+        }
```
cheehongw: perhaps instead of repeating the return string twice to handle the different status of the Task, you can first do something like:

```java
public String toString() {
        String statusString = super.getStatus() == 1 ? "[E][X]" : "[E][ ]";
        return statusString + super.getName() + " (at: "
                    + time.format(DateTimeFormatter.ofLocalizedDate(FormatStyle.FULL)) + ")\n";
        

```

what do you think 🤔 ?

src/main/java/Task.java
```diff
@@ -0,0 +1,38 @@
+import java.io.Serializable;
+
+class Task implements Serializable {
+
+    // 1 indicates done and 0 indicates not done
+    private int status;
```
cheehongw: why not use a boolean?

src/main/java/Duke.java
```diff
@@ -1,10 +1,223 @@
...
+            } else {
+                // If user input does not match any commands
+                System.out.println("invalid command! try 'help' for list of commands");
+            }
+            userInput = sc.nextLine();
+            split = userInput.split(" ");
+        }
+        System.out.println("goodbye!");
+        try {
+            FileOutputStream writer = new FileOutputStream("../../../data/data.txt");
```
cheehongw: Try to avoid "magic values" such as this `"../../../data/data.txt"` filepath here.

Perhaps declare it as a constant instead?

src/main/java/Duke.java
```diff
@@ -1,10 +1,223 @@
...
+                        System.out.printf("task removed:\n%s", t.toString());
+                        System.out.printf("you now have %d tasks\n", toDoList.size());
+                    } catch (IndexOutOfBoundsException e) {
+                        System.out.println(("the index you have entered does not exist!"));
+                    }
+                } catch (IndexOutOfBoundsException e) {
+                    System.out.println("delete description cannot be empty!");
+                } catch (NumberFormatException e) {
+                    System.out.println("delete description must be an integer!");
+                }
```
cheehongw: try not to nest try-catch within try-catch

cheehongw (commented):
The `main` method in `Duke.java` is too long and difficult to follow. It is deeply nested with try-catch and if-else blocks.

Maybe try and abstract out the error handling parts to other methods to reduce the number of try-catch blocks? 🤔

PR #184, opened 2022-01-25
[Zhang Zhikai] iP

zzhikai:
[](https://nus-cs2103-ay2122s2.github.io/website/schedule/week4/project.html#dukepro)DukePro
“Your mind is for having ideas, not holding them.” – David Allen ([source](https://dansilvestre.com/productivity-quotes))

### **Tasks**
- [x] ~~A-Gradle~~

- [x] ~~A-CheckStyle~~

- [x] ~~Level-10~~

- [ ] Fix Styling


src/main/java/Duke/Task/Deadline.java
```diff
@@ -0,0 +1,54 @@
...
+        this.by = by;
+        List<String> dateFormats = Arrays.asList("yyyy-MM-dd","dd-MM-yyyy", "dd/MM/yyyy", "dd:MM:yyyy");
+
+        DateTimeFormatter formatToParse;
+            for (int i = 0; i <= dateFormats.size(); i++) {
+                try {
+                    formatToParse = DateTimeFormatter.ofPattern(dateFormats.get(i));
+                    this.date = LocalDate.parse(by, formatToParse);
+                    break;
+                } catch (Exception wrongFormat) {
```
DavidTan0527: This will also capture the Exception at line 31 when `i == dateFormats.size()`.
Perhaps, it might be a better practice to put the `if` statement at line 36 before `try...catch`, and catch only `DateTimeParseException`.

src/main/java/Duke/Duke.java
```diff
@@ -0,0 +1,84 @@
...
+
+import Duke.Command.*;
+import Duke.Parser.Parser;
+import Duke.Storage.Storage;
+import Duke.Task.*;
+import Duke.Exception.*;
+import Duke.TaskList.TaskList;
+import Duke.Ui.Ui;
+
+import java.io.*;
```
DavidTan0527: Avoid wildcard imports. It might be better to explicitly import the packages you're using.

src/main/java/Duke/Parser/Parser.java
```diff
@@ -0,0 +1,59 @@
+package Duke.Parser;
+
+import Duke.Command.*;
```
DavidTan0527: Avoid wildcard imports. It might be better to explicitly import the packages you're using.

src/main/java/Duke/Storage/Storage.java
```diff
@@ -0,0 +1,47 @@
+package Duke.Storage;
+
+import Duke.Task.Task;
+
+import java.io.*;
```
DavidTan0527: Avoid wildcard imports. It might be better to explicitly import the packages you're using.

src/main/java/Duke/Task/Event.java
```diff
@@ -0,0 +1,23 @@
+package Duke.Task;
+
+/**
+ * Represents an event task.
+ */
+public class Event extends Task{
```
DavidTan0527:
```suggestion
public class Event extends Task {
```

src/main/java/Duke/TaskList/TaskList.java
```diff
@@ -0,0 +1,150 @@
+package Duke.TaskList;
+
+import Duke.Task.Task;
+import Duke.Exception.*;
```
DavidTan0527: Avoid wildcard imports.
```suggestion
import Duke.Exception.DukeException;
```

src/main/java/Duke/TaskList/TaskList.java
```diff
@@ -0,0 +1,150 @@
...
+            if (taskStore.get(i).toString().contains(searchWord)) {
+                searchResult.add(taskStore.get(i));
+            }
+        }
+        if (searchResult.size() == 0) {
+            throw new EmptyListException("No results found!");
+        }
+        for (int i = 0; i < searchResult.size(); i++) {
+            System.out.println((i + 1) + "." + searchResult.get(i).toString());
+        }
```
DavidTan0527: This is quite similar to `printOutTaskList`. Maybe we can create a `TaskList` here after filtering, then call the method as to not repeat the code.

DavidTan0527 (commented):
Overall great code! Just a few coding standard issues and some suggestions that you can hopefully look into. 😄

src/main/java/Duke/Task/Deadline.java
zzhikai: Thanks for the suggestion. 
What I was intending to do here is that because the LocalDate.parse will always return exception if the format used is incorrect, I needed to catch the Exception to continue to try the next format, only when all formats are exhausted, I throw the exception. 
Hence its kind of difficult to put the statement outside(I have to admit this solution is kind of hard coded as well)😅

src/main/java/Duke/TaskList/TaskList.java
zzhikai: Thank you for this suggestion! Makes the code a lot more efficient and clean👍  8475c55b322b6e5013a713391e12e80d7f5378e9

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,88 @@
...
+import java.io.FileOutputStream;
+import java.io.FileNotFoundException;
+import java.io.IOException;
+import java.io.ObjectOutputStream;
+
+
+public class Duke {
+
+    private static Ui ui;
+    private Ui uiDuke;
```
yusufaine: Was `uiDuke` used anywhere else?

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,88 @@
...
+    }
+
+    private static void printFileContents(String filePath) throws FileNotFoundException {
+        File f = new File(filePath);
+        Scanner s = new Scanner(f);
+        while (s.hasNext()) {
+            System.out.println(s.nextLine());
+        }
+
+    }
```
yusufaine: I'm curious, what was your rationale for placing this command here? 
I personally would consider moving this into `Command` or `Ui`?

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,88 @@
...
+        try {
+            FileOutputStream writeDatabaseInput = new FileOutputStream("src/main/java/TaskDatabase.ser");
+            ObjectOutputStream writeDatabaseStream = new ObjectOutputStream(writeDatabaseInput);
+            writeDatabaseStream.writeObject(inputTaskList);
+            writeDatabaseStream.flush();
+            writeDatabaseStream.close();
+        } catch (IOException e) {
+            e.printStackTrace();
+        }
+    }
```
yusufaine: Similar to `printFileContents()`

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,88 @@
...
+
+    public static int indexOfTask(String command) {
+        int index = command.length() + 1;
+        return index;
+    }
+
+    public static int endIndexOfTask(String input) {
+        int index = input.indexOf("/");
+        return index;
+    }
```
yusufaine: Consider adding JavaDocs to explain this portion as I don't quite understand what this methods are used for, sorry 😅

src/main/java/duke/command/DeleteCommand.java
```diff
@@ -0,0 +1,43 @@
...
+
+        if (description == "") {
+            throw new EmptyDescriptionException("OOPS!!! The description cannot be empty.");
+        }
+        try {
+            int taskNumber = Integer.valueOf(description);
+            taskIndex = taskNumber - 1;
+        } catch (NumberFormatException e) {
+            throw new DukeException("Please input number after unmark");
+        }
```
yusufaine: Errors that you may consider checking for:
1. Invalid number, and
2. Index out of bounds

src/main/java/duke/command/DeadlineCommand.java
```diff
@@ -0,0 +1,51 @@
...
+
+    /**
+     * Executes DeadlineCommand to add deadline task to a TaskList.
+     *
+     * @param taskList TaskList for command to add task to.
+     */
+    @Override
+    public void callCommand(TaskList taskList) {
+        taskList.addTask(this.task);
+    }
```
yusufaine: Not really commenting about your code here, but I like how you're handling the execution of the command, I might modify mine to make it like this as it makes more sense! 👍

src/main/java/duke/exception/EmptyByException.java
```diff
@@ -0,0 +1,10 @@
+package duke.exception;
```
yusufaine, line 1: I think this is a good way to handle exceptions as the class name is self-explanatory. 👍

src/main/java/duke/parser/Parser.java
```diff
@@ -0,0 +1,59 @@
...
+     */
+    public Command parse() throws DukeException {
+        // split command away
+        String[] inputStrings = this.userInputLine.trim().split(" ", 2);
+        String inputCommand = inputStrings[0];
+        // has command and body
+        String inputBody = inputStrings.length == 2 ? inputStrings[1] : "";
+        switch (inputCommand) {
+        case "list":
+            return new ListCommand("");
```
yusufaine, line 45: more nit-picky then anything, I find it a little odd a new object is initialised with an empty string  but don't worry, I understand that the `Command` parent class required you to give a parameter 😆

src/main/java/duke/storage/Storage.java
```diff
@@ -0,0 +1,53 @@
...
+    private static String filePath = "src/main/java/TaskDatabase.ser";
+
+    public Storage(String path) {
+        filePath = path;
+    }
+
+
+    //@@author zzhikai-reused
+    //Reused from https://samderlust.com/dev-blog/java/write-read-arraylist-object-file-java
+    // with minor modification
```
yusufaine, line 30: Was there any particular reason why you used his code instead of something simpler like `FileReader`/`FileWriter`? 🤔

src/main/java/duke/ui/Ui.java
```diff
@@ -0,0 +1,18 @@
+package duke.ui;
+
+public class Ui {
```
yusufaine, line 3: Maybe you can consider bringing the user input/output here since `UI`  is supposed to "deals with interactions with the user". I personally made all my print methods and `Scanner` in this class so that the I/O stuff is isolated here.

yusufaine (commented):
Overall rather clean code layout, just a few nit-picky stuff here and there so don't take them too seriously, it's mainly because we've done our code a little differently so I was curious as to why you did things a certain way and left some suggestions (that you can choose to (not) implement 😆 )

PR #185, opened 2022-01-25
[Loy Yoke Yue] iP

yokester98:
# **Duke Chatbot**
Duke helps you to schedule your things to do. It's
- Easy to use
- Fast
- ~~Pay-to-win~~ Completely **_FREE_**
> "Why spend money on schedulers when Duke is free to use?" ~ CS student probably

## Simply follow the instructions below to get Duke:
1. Download it from [here](https://github.com/yokester98/ip).
2. Run the program.
3. Add your tasks.
4. Sit back and relax as Duke organises your schedule for you! 😄 

Current features include:
- [x] Manage tasks
- [ ] Reminders (coming soon!)

```java
public class Main {
    }
}
```


src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,107 @@
...
+     * @return  ArrayList containing the Task objects.
+     * @throws IOException  If fail to read the file.
+     */
+    public ArrayList<Task> load() throws IOException {
+        ArrayList<Task> loadedTasks = new ArrayList<>();
+        FileReader fReader = new FileReader(String.valueOf(this.filePath));
+        BufferedReader reader = new BufferedReader(fReader);
+        String line;
+        while ((line = reader.readLine()) != null) {
+            String[] strArr;
```
Korochii, line 46: Might be a little nitpicking here but I think strArr should be in plural form since it is an array of Strings.

src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,107 @@
...
+            Task task = new Task("");
+            strArr = line.split(" \\| ");
+            switch (strArr[0]) {
+            case "T":
+                task = new Todo(strArr[2]);
+                if (strArr[1].equals("1")) {
```
Korochii:
```suggestion
                if (strArr[1].equals("X")) {
```
I think there is a bug in which marked to-do tasks are being loaded as unmarked to-do tasks upon re-running Duke (which might be caused by this line)

src/main/java/duke/Task.java
```diff
@@ -0,0 +1,86 @@
...
+    }
+
+    /**
+     * Returns a String of the description of the task.
+     *
+     * @return  description of the task
+     */
+    public String getDescription() {
+        return this.description;
+    }
```
Korochii:
```suggestion
    public String getDescription() {
        return description;
    }
```
Is it necessary to include `this` here? (since removing it does not make the code ambiguous)

src/main/java/duke/Task.java
```diff
@@ -0,0 +1,86 @@
...
+    public void setDone() {
+        this.isDone = true;
+    }
+
+    /**
+     * Sets the isDone attribute to false.
+     */
+    public void setNotDone() {
+        this.isDone = false;
+    }
```
Korochii: The implementation here works fine. However, my personal preference would be to merge these 2 methods into a single method that takes in a boolean value, since both methods are essentially doing the same thing (which is to set the value of isDone)

src/main/java/duke/Ui.java
```diff
@@ -0,0 +1,120 @@
...
+        System.out.println(LINE);
+        DukeException error;
+        if (type.equals("")) {
+            error = new DukeException();
+        } else {
+            error = new DukeException(type);
+        }
+        System.out.println(error);
+        System.out.println(LINE);
+    }
```
Korochii: Interesting approach in handling exceptions! (which I have not thought of...)

Korochii (commented):
Overall, your code seems quite clear and readable, and had very little coding standard/quality violations as well (which was why I had to nitpick a few areas :sweat_smile:)

Good job and keep up the good work!

src/main/java/duke/Storage.java
yokester98: You are right! Thanks so much for noticing this! Fixed the issue in latest commit.

PR #188, opened 2022-01-25
[Sim Jun Heng] iP

simjunheng:
# Duke2k22
> “Life is too complicated not to be orderly.” 😀 – Martha Stewart [(source)](https://dansilvestre.com/productivity-quotes/)

## Why should you use Duke2k22
- text-based
- easy to learn
- ~~FAST~~ *ULTRA FAST* to use

## How to get started
1. `download` from [here](https://github.com/simjunheng/ip/releases/tag/A-Jar)
2. `double-click` it 
3. `add` your tasks
4. sit back and relax 😘 

Not to mention, it's **FREE**

## Key Features
```ruby
- [x] Manage Tasks
- [ ] Add Reminders(coming soon)
```


src/main/java/duke/Deadline.java
```diff
@@ -0,0 +1,58 @@
...
+public class Deadline extends Task {
+
+    private LocalDate date;
+
+    /**
+     * Assigns desc, date and done to this instance.
+     *
+     * @param desc the task description.
+     * @param date the date of the event.
+     * @param isDone the current completion status of the task.
```
kxshxsh, line 21: Really appreciate how your javadocs follow conventions to a T!

src/main/java/duke/Event.java
```diff
@@ -0,0 +1,59 @@
...
+        super(desc, isDone);
+        this.date = date;
+    }
+
+    @Override
+    public String toString() {
+        String str = "[E]";
+        // Convert Current Date Format to MMM dd yyyy
+        String customPattern = "MMM dd yyyy";
+        DateTimeFormatter customFormat = DateTimeFormatter.ofPattern(customPattern);
```
kxshxsh, line 38: Just a suggestion, but in my opinion, I think you can skip line 31 and directly alter line 32 to be `DateTimeFormatter customFormat = DateTimeFormatter.ofPattern("MMM dd yyyy");`

I don't think it would break the logical flow or make it less comprehensive. Maybe this is something you can consider if you agree :)

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,58 @@
...
+     * @param str the string entered by the user.
+     */
+    public Parser(String str) {
+        String[] temp = str.split(" ");
+        command = temp[0];
+        if (temp.length > 1) {
+            int index = 1;
+            while (index < temp.length) {
+                if (!temp[index].equals("/by") && !temp[index].equals("/at")) {
+                    if (desc == null) {
```
kxshxsh: An alternative approach could be to use `str.indexOf(...)` to search for "/by" and "/at" and then use `str.substring(...)` to help extract the different sections of the input string. I think it could help save on the while loops and also help improve the readability of your code. What do you think?

src/main/java/duke/TaskList.java
```diff
@@ -0,0 +1,197 @@
...
+            System.out.println("NOTED. I'VE REMOVED THIS TASK:");
+            System.out.println(task);
+            System.out.println("NOW YOU HAVE " + list.size() + " TASKS IN THE LIST.");
+        }
+    }
+
+    /**
+     * Adds to do task into our current list.
+     *
+     * @param desc the task description entered by the user
```
kxshxsh: Something small that you have been doing consistently otherwise but have seemed to miss out here is the period after the parameter description in your javadocs. Very tiny detail but something good to be conscious about:)

src/main/java/duke/TaskList.java
```diff
@@ -0,0 +1,197 @@
...
+    }
+
+    /**
+     * Finds a list of tasks with a keyword specified by the user.
+     *
+     * @param taskName the keyword entered by the user.
+     */
+    public void findTask(String taskName) {
+        int num = 1;
+        int matchNum = 0;
```
kxshxsh: Subjectively, I think `matchNum` can be named something clearer, more descriptive and detailed:) Such as `numOfMatches`

src/main/java/duke/Ui.java
```diff
@@ -0,0 +1,90 @@
...
+        Storage.writeFile(list);
+        System.out.println("Bye. Hope to see you again soon!");
+    }
+
+    /**
+     * Handles invalid inputs.
+     *
+     * @throws DukeException if invalid inputs are detected.
+     */
+    public static void notSpecified() throws DukeException {
```
kxshxsh: Our coding standard says that "Names representing methods must be verbs and written in camelCase." Perhaps you can consider changing the name of `notSpecified()` to fit the conventions better?

kxshxsh (commented):
Hi!! I think you have done a great job in your project:) Everything seems well thought out and elaborate! Your javadocs are also near flawless! There are, however, some instances where I think there is an easier alternative so I have pointed them out for you to consider and evaluate!

src/main/java/duke/Deadline.java
```diff
@@ -0,0 +1,58 @@
...
+        String str = "[D]";
+        // Convert Current Date Format to MMM dd yyyy
+        String customPattern = "MMM dd yyyy";
+        DateTimeFormatter customFormat = DateTimeFormatter.ofPattern(customPattern);
+        String tempDate = customFormat.format(this.date);
+        if (super.isDone) {
+            str += "[X] " + super.desc + " (by: " + tempDate + ")";
+        } else {
+            str += "[ ] " + super.desc + " (by: " + tempDate + ")";
+        }
```
ryanwalterlee, line 44: In my opinion, I prefer to write strings in java using template literals, such as 
```
String.format("[X] %s by: %s)", super.desc, tempDate);
```
This way, it is not as easy to miss out a " " (space), which could lead to bugs that could take hours to resolve. If this syntax interests you, you could try this next time :)

src/main/java/duke/Event.java
```diff
@@ -0,0 +1,59 @@
...
+
+    private LocalDate date;
+
+    /**
+     * Assigns desc, date and done to this instance.
+     *
+     * @param desc the task description
+     * @param date the date of the event
+     * @param isDone the current completion status of the task.
+     */
```
ryanwalterlee: According to the coding standard stated here [link](https://se-education.org/guides/conventions/java/intermediate.html), it states that all javadoc should have the following form where each @params line ends with a fullstop. IsDone has a fullstop at the end so perhaps you overlooked the previous 2.

src/main/java/duke/TaskList.java
```diff
@@ -0,0 +1,197 @@
...
+     */
+    public void addEvent(String desc, String date, boolean done) throws DukeException {
+        if (desc == null) {
+            throw new DukeException("PLEASE INCLUDE TASK DESCRIPTION IN YOUR COMMAND.");
+        }
+        if (date == null) {
+            throw new DukeException("PLEASE INCLUDE THE DATE IN YOUR COMMAND.");
+        }
+        try {
+            LocalDate date2 = LocalDate.parse(date);
```
ryanwalterlee: I think the variable name 'date2' is not very descriptive of what it is. At first look, I don't exactly know what date2 is. After inspecting the code, I realised that date2 is actually a parsed date. Perhaps the variable name could be called dateParsed instead? I understand variable naming is difficult when there are more important logic to handle but every small effort goes a long way for the readability of code.

ryanwalterlee (commented):
Overall good coding standard observed, had a hard time trying to look for violations.

From what I see throughout, some things to take note of:
1. javadoc comments ends with '.'
2. variable naming could be more descriptive (try thinking in a strangers shoes to determine if someone else can understand)

src/main/java/duke/Deadline.java
s7manth: I agree with this. 

In fact, I created a custom `OutputFormatter` just for a bulk `append` method for `StringBuilder`.

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,53 @@
...
+ * @version CS2103T AY21/22 Sem 2
+ */
+public class Duke {
+
+    private TaskList list = new TaskList();
+    public Duke() throws IOException, DukeException {
+        list = Storage.loadFile();
+    }
+
+    public String getResponse(String input) throws IOException {
```
s7manth: Instead of using so many if-else statements, you could have abstracted out the handling commands part. `Parser` identifies the type of command, and maybe `Ui` could have handled the command.

src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,104 @@
...
+
+    /**
+     * Loads task data from file and adds it into a list.
+     * If file or directory is not found, the respective items are created and
+     * an empty array list is returned.
+     *
+     * @return an arraylist containing data read form the file.
+     * @throws IOException if there's error arising from the file methods
+     * @throws DukeException if data in the file is corrupted (not in the correct storage format)
+     */
```
s7manth: Great Java Docs!

src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,104 @@
...
+    }
+
+    /**
+     * Writes data from the arraylist into file.
+     *
+     * @param list an arraylist containing task objects.
+     * @throws IOException if there's error arising from I/O methods.
+     */
+    public static void writeFile(TaskList list) throws IOException {
+        FileWriter fw = new FileWriter("./data/duke.txt");
```
s7manth: Should create the directory and the file, if they didn't exist to avoid `FileNotFoundException`.

s7manth (commented):
Overall, the code was of good quality. LGTM

PR #193, opened 2022-01-25
[Shem] iP

sheimoria:
# Duke (Shem's Version)

> “It is only with the heart that one can see rightly; what is essential is invisible to the eye.” – Antoine de Saint-Exupéry,  [The Little Prince](https://en.wikipedia.org/wiki/The_Little_Prince)

*Daddy* Duke frees your mind of having to remember all the things you need to do. It's

- Text-based
- Easy to learn
- ~~Fast~~ Super fast to use
- Completely **free**!

## Oh my days how do I get it?

1. Download it from [here](https://github.com/nus-cs2103-AY2122S2/ip/pull/193).
2. Double-click it
3. Add your tasks
4. Let it manage your tasks for you 😉

## What can this baby do?

- [ ] Manage to-dos, deadlines and events
- [ ] Find tasks with a specific keyword

If you're a Java programmer, you can use it to practice Java too. Here's a snippet of the `main` method:

```java
public void run() {
     ui.showWelcome();
     boolean isExit = false;
     while (!isExit) {
         try {
             String fullCommand = ui.readCommand();
             Command c = Parser.parse(fullCommand);
             c.execute(storage, tasks, ui);
             isExit = c.isExit();
         } catch (DukeException e) {
             ui.showError(e.getMessage());
         }
     }
 }

 }
```


src/main/java/Duke.java
```diff
@@ -1,10 +1,17 @@
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Scanner;
 
-
 public class Duke {
     private static ArrayList<Task> tasks = new ArrayList<>();
 
+    private static String isPlural(int n) {
```
geraldneo567: I think it may be better if this method is renamed, since it may be confused with one that returns a boolean true or false value.

src/main/java/Duke.java
```diff
@@ -28,11 +28,11 @@ private static void load() throws IOException, DukeException {
             switch(type) {
```
geraldneo567: Should have a space before the bracket i.e `switch (type)`. Just a minor code style thing though

src/main/java/Duke.java
```diff
@@ -28,11 +28,11 @@ private static void load() throws IOException, DukeException {
...
                 case "E":
                     String at = commandArr[3];
-                    tasks.add(new Event(description, status, at));
+                    tasks.add(new Event(description, at, status));
                     break;
                 case "D":
```
geraldneo567: I believe the coding standard states no indentation for case clauses in switch cases. Should not affect the code though!

src/test/java/duke/TaskListTest.java
```diff
@@ -0,0 +1,16 @@
+package duke;
+
+import duke.task.ToDo;
+import org.junit.jupiter.api.Test;
+
```
geraldneo567: I believe the static imports should be at the top based on coding standard (although not 100% sure)

PR #194, opened 2022-01-25
[Tan Jie Wei] iP

jiewei98:
# **Duke**
> "Believe You Can and You're Halfway There" - Theodore Roosevelt ([Source](https://www.motivateamazebegreat.com/2017/07/believe-you-can-and-youre-halfway-there-meaning.html))
 
Duke is a task reminder application that has the following applications:
- Task creation (can be tagged as a To-do, Event or Deadline)
- Auto-saving of tasks when tasks are updated
- Searching of tasks by keyword

All you need to do is,
1. download it from [here](https://github.com/jiewei98/ip) 
2. run the Duke.main() file
3. and add your tasks!

Features:
- [X] Managing tasks
- [ ] Managing deadlines (To be added)
- [ ] Reminders (To be added)

Here is the `main` method in the Duke.main() file:
```ruby

        Ui ui = new Ui();
        ui.displayGreeting();

        Storage storage = new Storage();
        ArrayList<Task> list = storage.load();

        Parser parser = new Parser();
        CommandList commandList = new CommandList();

        while (!parser.isTerminated) {
            try {
                commandList.execute(parser.readInput(), parser, list, storage);
            } catch (IOException e) {
                System.out.println("Unable to read input!");
            }
        }
    }
```

Thanks for the read! :grin:


src/main/java/Duke/Storage.java
```diff
@@ -0,0 +1,115 @@
...
+    public ArrayList<Task> load() {
+        ArrayList<Task> list = new ArrayList<>(); // init arraylist
+
+        try {
+            FileReader fr = new FileReader(this.filePath.toString());
+            BufferedReader br = new BufferedReader(fr);
+            StringBuffer sb = new StringBuffer();
+            String currLine;
+
+            while ((currLine = br.readLine()) != null) {
```
Graphcalibur, line 56: This code block is a little difficult to read. Maybe include some line breaks between each condition?

src/main/java/Duke/Ui.java
```diff
@@ -0,0 +1,12 @@
+package Duke;
+
+public class Ui {
```
Graphcalibur: Does this need to be a class of its own?

src/main/java/Duke/CommandList.java
```diff
@@ -0,0 +1,133 @@
...
+
+    /**
+     * Execute different actions based on user's input taken from Parser.
+     *
+     * @param input command to execute
+     * @param parser Parser object to terminate program if necessary
+     * @param list Arraylist of Tasks to manipulate
+     * @param storage Storage object to load/store list of Tasks
+     */
+    public void execute(String input, Parser parser, ArrayList<Task> list, Storage storage) {
```
Graphcalibur: I find this function to be long and cumbersome to read. Perhaps you could include some line breaks, or extract some parts of the code into a separate function?

src/main/java/Duke/CommandList.java
```diff
@@ -0,0 +1,133 @@
...
+
+        if (input.equals("bye")) { // check for exit condition
+            System.out.println("Bye. Hope to see you again soon!");
+            parser.terminate();
+        } else if (input.equals("list")) { // list all tasks
+            System.out.println("Here are the tasks in your list:");
+            for (int i = 0; i < list.size(); i++) {
+                System.out.println((i + 1) + "." + list.get(i));
+            }
+        } else if (firstWord.equals("mark") || firstWord.equals("unmark")) { // mark/unmark task
```
Graphcalibur: It seems like it's possible for this part of the code to throw an exception (such as if secondword can't be converted to an int or if index goes outside the bounds of the ArrayList). Perhaps you can catch and handle these exceptions?

src/main/java/Duke/CommandList.java
```diff
@@ -0,0 +1,133 @@
...
+                    System.out.println("Got it. I've added this task:\n"
+                            + event + "\nNow you have " + list.size() + " tasks in your list.");
+                    storage.store(list);
+                } catch (ParseException e) {
+                    System.out.println("Please enter a valid date!");
+                }
+            }
+        } else if (firstWord.equals("find")) { // find tasks by a keyword
+            String secondWord = inputs[1];
+            List<Task> matches = new ArrayList<>();
```
Graphcalibur: Is there a reason why matches is of type List<> and not ArrayList<>?

src/main/java/Duke/CommandList.java
```diff
@@ -0,0 +1,133 @@
...
+                } catch (ParseException e) {
+                    System.out.println("Please enter a valid date!");
+                }
+            } else { // adds an event task
+                if (inputs.length == 1) {
+                    System.out.println("☹ OOPS!!! The description of an event cannot be empty.");
+                    parser.terminate();
+                }
+                try {
+                    SimpleDateFormat dateFormatter = new SimpleDateFormat("dd/MM/yyyy HHmm");
```
Graphcalibur: There's some copy-pasted code here with parsing and formatting the date. Maybe you could extract this out into a function to avoid code duplication?

Graphcalibur (commented):
Overall, your code mostly seems to follow good practices, but there are a few issues I have that I highlighted in this review.

PR #196, opened 2022-01-25
[Louis Lie] iP

louisdavinlie:
## Duke

> "If you don't have an assistant, you are one!" - Cameron Herold

Duke is the only personal assistant you ever need. It's,

- Interactive
- Easy to learn
- Text-based
- ~~Fast and easy~~ **_SUPER FAST AND EASY_** to use

To begin using this amazing assistant, you need to:

1. Download the code base from [here](https://github.com/louisdavinlie/ip) ⬇️ 
2. Run the Launcher class 🚀 
3. Add your tasks 📃 
4. Let it make your life simpler! 🤑 

Features:

- [x] Add any task you have.
- [x] Manage your tasks, deadlines and events.
- [x] Find any task you have.
- [ ] Reminders! (Coming soon)
----------------
Below is the `main` method in the Launcher class you need to run,

```
public class Launcher {
    }
}
```


src/main/java/Deadline.java
```diff
@@ -0,0 +1,13 @@
...
+    protected String dateAndTime;
+
+    public Deadline(String description, String dateAndTime) {
+        super(description);
+        this.dateAndTime = dateAndTime;
+    }
+
+    @Override
+    public String toString() {
+        return "[D]" + this.getStatusIcon() + " " + super.description + "(by:" + this.dateAndTime + ")";
```
lkldev: Following the DRY Principle, instead of using `this.getStatusIcon() + " " + super.description`, I feel that it will be better to use the toString method in the parent class `Task`, since the code is repeated for all 3 types of tasks.
```suggestion
        return "[D]" + super.toString() + "(by:" + this.dateAndTime + ")";
```

src/main/java/Duke.java
```diff
@@ -1,10 +1,153 @@
...
+                if (instructions.length == 1) {
+                    throw new DukeException("Please specify the description and time of the " + command + " you want to add.");
+                }
+                String[] taskAndTime = instructions[1].split("/");
+                if (taskAndTime.length < 2 || taskAndTime[0].equals("") || taskAndTime[1].length() < 3) {
+                    throw new DukeException("Invalid input. Please specify the description/time of the " + command + ".");
+                }
+            } else {
+                throw new DukeException("I don't understand your query.");
+            }
```
lkldev: I feel that using a switch case here would be neater and easier to read. You can use the "fall through" method to process the same type of command, e.g.
```java
switch (command) {
...
case "mark":
//Fall through
case "unmark":
//Fall through
case "delete":
    if (instructions.length == 1) {
            throw new DukeException("Please specify the task number to " + command + ".");
    }
    int taskNumber = Integer.parseInt(instructions[1]);
    if (taskNumber > taskList.size() || taskNumber <= 0) {
        throw new DukeException("Invalid task number! You have " + taskList.size() + " task(s).");
    }
    break;
...
}
```

src/main/java/Storage.java
```diff
@@ -0,0 +1,73 @@
...
+            this.file = new File(this.dataPath);
+            if (file.createNewFile()) {
+                System.out.println("File is not found. A new file has been created for you.");
+            } else {
+                System.out.println("File is found.");
+            }
+        } catch (IOException e) {
+            System.out.println(e.getMessage());
+        }
+    }
```
lkldev: Great method to abstract out the file checking process! Makes the code more readable 👍

src/main/java/Duke.java
```diff
@@ -1,10 +1,153 @@
...
+    public static void greetings() {
+        String logo = "";
+        System.out.println(HORIZONTAL_LINE + "\nHi! I'm Duke\nWhat can I do for you?");
+    }
+
+    public static void add(String[] instructions) {
+        Task task = null;
+        if (instructions[0].equals("todo")) {
+            task = new Todo(instructions[1]);
+            System.out.println("Added a to do task.");
```
lkldev: I feel that all the printed messages can be replaced with a named constant so that it will be easier to read the code since it is not important as to what the bot would print. An example would be:
```java
private static final String ADD_TODO_MESSAGE = "Added a to do task.";
...
System.out.println(ADD_TODO_MESSAGE);
```
This is easily understood that the bot would print a message. Also, having all the messages consolidated allow for easy changing of the message sent if an update is required.

src/main/java/Duke.java
```diff
@@ -1,10 +1,153 @@
...
+                System.out.println(e.getMessage());
+            }
+        }
+    }
+
+    public static void main(String[] args) throws IOException {
+        greetings();
+        chat();
+        bye();
     }
```
lkldev: Great level of abstraction here! it follows the Single Level of Abstraction Principle, making it easy to understand what is the main logic here.

src/main/java/Task.java
```diff
@@ -0,0 +1,26 @@
...
+        return isDone ? "[X]" : "[ ]";
+    }
+
+    public void markAsDone() {
+        this.isDone = true;
+    }
+
+    public void unmark() {
+        this.isDone = false;
+    }
```
lkldev: Since there is a `markAsDone` method, I feel that this method should be named as `unmarkAsDone` or something similar so that it is easy to understand that these two methods are opposites of each other.

src/main/java/Duke.java
```diff
@@ -1,10 +1,153 @@
...
+                if (instructions[0].equals("bye")) {
+                    storage.writeToFile(taskList);
+                    break;
+                } else if (instructions[0].equals("list")) {
+                    list();
+                } else if (instructions[0].equals("mark") || instructions[0].equals("unmark") || instructions[0].equals("delete")) {
+                    modify(instructions);
+                } else {
+                    add(instructions);
+                }
```
lkldev: Following the Single Level of Abstraction Principle, since there is a `checkCommand` method, maybe it will be good to have a `processCommand` method as well?

lkldev (commented):
Overall, good code! There are some nits that I have picked out based on the [code quality guidelines](https://nus-cs2103-ay2122s2.github.io/website/se-book-adapted/chapters/codeQuality.html). Generally, great naming of variables and methods, and the code is easy to follow.

PR #204, opened 2022-01-25
[HO QI YUAN JANALD] iP

janald99:
### **DUKE**

> "Tasks are tasks. You gotta do what you gotta do. But *automating* does what you gotta do for you." - *Janald Ho*

Duke helps you keep track of your tasks at hand, so that you don't have to worry about forgetting what you need to do. It's,
- text-based
- *easy* to use ☺️ 
-  ~~FAST~~ **SUPER FAST** to use 😍 
- most importantly, **__FREE__**!!! 🤑 

All you need to do is,

1. download it from [here](https://github.com/janald99/ip).
2. double-click it. 
3. add your tasks.

Features:

- [x] Managing tasks
- [ ] Reminders (coming soon)

If you are a Java programmer, you can use it to practice Java too. Here's the ```main```  method:
```ruby
public class Main {
    public static void Main (String[] args) {
    }
}
```


src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,57 @@
...
+    /**
+     * Runs the chatbot application.
+     */
+    public void run() {
+        ui.greet();
+        boolean isExit = false;
+        while(!isExit) {
+            try {
+                  String fullCommand = ui.readCommand();
+                  ui.divide();    // show the divider line ("__________")
```
blaketingyu: I like the way you use comments to make your code more readable!

src/main/java/duke/parser/Action.java
```diff
@@ -0,0 +1,16 @@
+package duke.parser;
+
+/**
+ * Actions according to the available commands in Duke.
+ */
+public enum Action {
```
blaketingyu, line 6: Good use of enum here! :D

src/main/java/duke/parser/Parser.java
```diff
@@ -0,0 +1,134 @@
+package duke.parser;
+
+import duke.command.AddCommand;
```
blaketingyu, line 3: I like how you listed the imports lexicographically! 👍

src/main/java/duke/task/Deadline.java
```diff
@@ -0,0 +1,54 @@
+package duke.task;
+
+import java.time.format.DateTimeFormatter;
+import java.time.LocalDateTime;
+
+/**
+ * A type of Task, which has a deadline the task is due by.
+ */
+public class Deadline extends Task {
+
```
blaketingyu, line 10: Maybe you could give each of your class attributes a short description. 👍

src/main/java/duke/task/Deadline.java
```diff
@@ -0,0 +1,54 @@
...
+    protected String by;
+    private LocalDateTime dateTime;
+
```
blaketingyu, line 13: Maybe you could give your Deadline constructor the method header! 👍

src/main/java/duke/task/Event.java
```diff
@@ -0,0 +1,28 @@
+package duke.task;
+
+/**
+ * A type of Task, which is to be done at somewhere.
+ */
+public class Event extends Task{
+    protected String at;
+
```
blaketingyu, line 8: Same as the deadline class, I think you could give your constructor a method header hehe 👍

src/main/java/duke/ui/Messages.java
```diff
@@ -0,0 +1,70 @@
+package duke.ui;
+
+/**
+ * A class that consists of messages that the Ui will show to the user as needed.
+ */
+public class Messages {
+
+    public static final String LINE_SEPARATOR = System.lineSeparator();
```
blaketingyu, line 8: I like how you use CAPS for the final variables! 👍

src/test/duke/parser/ParserTest.java
```diff
@@ -0,0 +1,46 @@
...
+import java.util.Arrays;
+
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.*;
+
+class ParserTest {
+
+    @Test
+    void testParse() {
```
blaketingyu: LGTM, I believe you've covered a lot of testing to me!

blaketingyu (commented):
LGTM overall, maybe just a feedback would be to give your methods their respective method headers. but I believe you just forgotten to comment for some of them. Nevertheless, great work! 👍

src/main/java/duke/task/Deadline.java
janald99: Thank you for your feedback! Would definitely add more descriptions in the next commits. 👍

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,57 @@
...
+ */
+public class Duke {
+
+    private static Storage storage;
+    private static TaskList tasks;
+    private static Ui ui;
+
+    public Duke(String filePath) {
+        this.ui = new Ui();
+        this.storage = new Storage(filePath);
```
MechFroG88, line 29: I think that `this.` can be omitted in this case since it is unnecessary.

src/main/java/duke/command/AddCommand.java
```diff
@@ -0,0 +1,86 @@
...
+                    tasks.add(new Todo(description));
+                    ui.printTaskAdded(tasks);
+                    storage.save(tasks);
+                } catch (InvalidArgumentException e) {
+                    ui.showError(e.getMessage());
+                } catch (IOException e) {
+                    ui.showError(Messages.SAVE_ERROR);
+                } finally {
+                    break;
+                }
```
MechFroG88: I think it would be better if you put this whole chunk as another function, so you would not have a very long switch case. (Since there might be more cases in the future)

src/main/java/duke/ui/Messages.java
```diff
@@ -0,0 +1,70 @@
+package duke.ui;
+
+/**
+ * A class that consists of messages that the Ui will show to the user as needed.
+ */
+public class Messages {
```
MechFroG88, line 6: I like your idea of putting all constant strings into a class!

src/main/java/duke/command/MarkCommand.java
```diff
@@ -0,0 +1,50 @@
...
+        this.inputWords = inputWords;
+    }
+
+    /**
+     * Marks the task as done.
+     *
+     * @param tasks TaskList that command is executed on.
+     *
+     * @param ui User interface that interacts with the user.
+     *
```
MechFroG88: Based on the javadocs guide in module website, I think there shouldn't be new line between @param.

MechFroG88 (commented):
Overall I think your code is very nicely written and neat!

Just some suggestions that you could possibly look into it.

src/main/java/duke/command/MarkCommand.java
janald99: i agree with you! Thank you, will fix that!

src/main/java/duke/command/AddCommand.java
janald99: Great idea 👍  will do it when I have the time!

PR #205, opened 2022-01-25
[Tan Chin Kiat] iP

tanchinkiat99:
Mr Tutu helps you remember all your tasks so you don't have to. It is

- easy to use
- quick to download
- and ~~$29.99/mth~~ _completely_ **FREE**

All you need to do is:

1. download it [here](https://github.com/nus-cs2103-AY2122S2/ip/pull/205),
2. add your tasks,
3. and let it manage your tasks for you! 😃 

Features:

- [x] Managing tasks
- [ ]  Managing deadlines (coming soon!)
- [ ] Reminders

If you Java programmer, you can use it to practice Java too. Here's a snippet of the `main` method:

``` ruby
public static void main(String[] args) throws InvalidInputException, IOException {
        Storage store = new Storage("./data");
        File f = store.initialize();

        Tutu.separator();
        System.out.println("Hello! I'm Tutu\nWhat can I do for you?");
        Tutu.separator();

        Ui ui = new Ui();
        String cmd = ui.read();
        while (!cmd.equals("bye")) {
            ...
        }
        System.out.println("Bye. Hope to see you again soon!");
    }
```


src/main/java/tutu/Tutu.java
```diff
@@ -0,0 +1,90 @@
+package tutu;
+
+import java.io.File;
+import java.io.IOException;
+import tutu.data.Storage;
+import tutu.data.TaskList;
+import tutu.exception.InvalidInputException;
+import tutu.task.*;
```
atmh: Wildcard import should not be used. To stop IntelliJ from doing it automatically, you could change the settings here
`File > Settings > Editor > Code Style > Java > Class count to use import with '*':`

src/main/java/tutu/data/Storage.java
```diff
@@ -0,0 +1,56 @@
...
+import java.io.File;
+import java.io.FileNotFoundException;
+import java.io.FileWriter;
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.Scanner;
+
+public class Storage {
```
atmh, line 12: I like that you have included Javadocs for most of your classes and methods 😄 . 
Just a few more Javadocs to include in Storage.java, Parser.java and UI.java 👍.

src/main/java/tutu/user/Parser.java
```diff
@@ -0,0 +1,32 @@
+package tutu.user;
+
+import tutu.exception.InvalidInputException;
+
+public class Parser {
+
+    public enum Function {
+        LIST, MARK, UNMARK, TODO, DEADLINE, EVENT, DELETE, FIND
+    }
```
atmh: I like that you included enum into your code 👍

src/main/java/tutu/user/Parser.java
```diff
@@ -0,0 +1,32 @@
...
+            return Function.DEADLINE;
+        } else if (cmd.length() > 4 && cmd.substring(0, 5).equals("event")) {
+            return Function.EVENT;
+        } else if (cmd.length() > 5 && cmd.substring(0, 6).equals("delete")) {
+            return Function.DELETE;
+        } else if (cmd.length() > 4 && cmd.substring(0, 4).equals("find")) {
+            return Function.FIND;
+        } else {
+            throw new InvalidInputException("☹ OOPS!!! I'm sorry, but I don't know what that means :-(");
+        }
```
atmh: Perhaps a switch statement would be neater. You could also consider splitting instead of substring. E.g.
```java
String[] splitSpace = input.split(" ");
String cmd = splitSpace[0];

switch (cmd) {
case "mark":
    ...
    break;
default:
    ...
}
```

src/main/java/tutu/Tutu.java
```diff
@@ -0,0 +1,90 @@
...
+                        break;
+                    case DELETE:
+                        taskList.delete(cmd);
+                        store.update(taskList);
+                        break;
+                    case FIND:
+                        taskList.find(cmd);
+                        break;
+                    default:
+                        throw new InvalidInputException("☹ OOPS!!! I'm sorry, but I don't know what that means :-(");
```
atmh: For the case and default clauses in switch statements, there should be no indentations. E.g.
```java
switch (condition) {
case ABC:
    statements;
    break;
default:
    statements;
    break;
}
```

atmh (commented):
Looks good 😄. 
Just some minor changes to be made regarding the coding standard 👍.

src/main/java/tutu/data/TaskList.java
```diff
@@ -0,0 +1,116 @@
...
+     */
+    public int taskNumber() {
+        return items.size();
+    }
+
+    /**
+     * Retrieves the task at a specified index.
+     * @param i Index of task to be retrieved.
+     * @return Task object representing the task at index i.
+     */
```
1001mei, line 51: Perhaps adding an empty line between description and parameter section is better according to Java coding standard? I noticed this issue in several other places too.
```suggestion
    /**
     * Retrieves the task at a specified index.
     *
     * @param i Index of task to be retrieved.
     * @return Task object representing the task at index i.
     */
```

src/main/java/tutu/data/TaskList.java
```diff
@@ -0,0 +1,116 @@
...
+    public TaskList() {
+        this.items = new ArrayList<Task>();
+    }
+
+    /**
+     * Returns the number of tasks in the TaskList object.
+     * @return Number of tasks.
+     */
+    public int taskNumber() {
```
1001mei, line 43: Method name would be better to start with verbs. For example, consider name such as `getNumberOfTasks`. What do you think?

src/main/java/tutu/task/ToDo.java
```diff
@@ -0,0 +1,27 @@
...
+/** Represents a ToDo task. */
+public class ToDo extends Task {
+    /**
+     * Constructor to create a ToDo object
+     * @param task Command input from user.
+     * @throws InvalidInputException If input is invalid.
+     */
+    public ToDo(String task) throws InvalidInputException {
+        super(task);
+        if (super.task.length() < 6) {
```
1001mei: I feel like using 6 as a magic number could be avoided. Perhaps you can consider introducing a constant variable like `private final int MINIMUM_DESCRIPTION_LENGTH = 6`?

src/main/java/tutu/task/Task.java
```diff
@@ -0,0 +1,49 @@
+package tutu.task;
+
+/** Represents a task. */
+public class Task {
+    /** Description of task */
+    protected String task;
```
1001mei, line 10: I guess using name like `taskDescription` here would be more precise? Maybe it's just me, but `task` somehow gives me a feel that it is an instance of class `Task`.

src/main/java/tutu/user/Parser.java
1001mei: I agree with @atmh, using enum makes the code more clean and concise. 👍

1001mei (commented):
Overall, I found your code easy to read and I am able to see the class structures in it without much effort. Keep up the good work! There are some coding standard violations though, maybe you can consider using [checkstyle](https://se-education.org/guides/tutorials/checkstyle.html) to help you resolve them.

src/main/java/tutu/Tutu.java
```diff
@@ -0,0 +1,90 @@
...
+
+/** Represents a todo list that takes in command inputs from the user. */
+public class Tutu {
+    /** TaskList object that stores the tasks added */
+    private static TaskList taskList = new TaskList();
+
+    /**
+     * Prints a line of asterisks to separate between user inputs and resulting outputs.
+     */
+    public static void separator() {
```
louisdavinlie: According to the Java coding standard, names representing methods must be verbs. Perhaps you can add a verb to the prefix of your current method name such as print.
```suggestion
    public static void printSeparator() {
```

src/main/java/tutu/task/Deadline.java
```diff
@@ -0,0 +1,48 @@
...
+            }
+        }
+    }
+
+    /**
+     * Returns a description of the task, whether it is done and the deadline.
+     * @return Description of task and deadline.
+     */
+    @Override
+    public String isDone() {
```
louisdavinlie, line 57: Do you think it would be better if the name of this method is toString? I feel that it would be more proper since this method returns the String representation of the task. I think isDone sounds more like a name for Boolean variables/methods.
```suggestion
    public String toString() {
```

src/main/java/tutu/task/Task.java
```diff
@@ -0,0 +1,49 @@
...
+    /** Boolean to represent if the task is done */
+    protected boolean done;
```
louisdavinlie, line 12: Consider adding "is" before the variable name to make it sound more like a Boolean variable, according to the Java coding standard.
```suggestion
    protected boolean isDone;
```

louisdavinlie (commented):
Looks excellent! The code in general is easy to read and follow

PR #207, opened 2022-01-26
[Ong Weng Qi] iP

chriswengqi:
[](https://nus-cs2103-ay2122s2.github.io/website/schedule/week4/project.html#dukepro)DukePro
“Never put off till tomorrow what may be done day after tomorrow just as well.” – Mark Twain [(source)](http://www.twainquotes.com/Procrastination.html)

DukePro is an ultimate procrastination killer which helps you remember what to do so that you cannot use this as an excuse to skip your tasks! It's,

- text-based
- easy to learn
- FAST SUPER FAST to use

All you need to do is,

1. double-click it.
1. add your tasks.
1. let it manage your tasks for you 😉

And it is **FREE**!

Features:

 

- [x] Managing tasks
- [ ]  Managing deadlines (coming soon)
- [ ]  Reminders (coming soon)

```
public class Main {
    }
}
```


src/main/java/Duke.java
```diff
@@ -1,10 +1,152 @@
...
+            }
+        } catch (FileNotFoundException e) {
+            ui.reply("Please create the text file data/duke.txt");
+        }
+
+
+
+    }
+
+    public void run() {
```
brigittesantoso, line 50: Avoid long methods. Try to split this method up into smaller parts.
image

src/main/java/Duke.java
```diff
@@ -1,10 +1,152 @@
...
+        ui.greet();
+
+        String tab = "    ";
+
+        parser.parse(sc.nextLine());
+        String task = parser.getTask();
+        String item = parser.getItem();
+
+
+        while (!task.equals("bye")) {
```
brigittesantoso: Each string here can be split into separate methods to improve readability of the code. 

Example:
public void deadlineTask() {
         if (!item.equals("")) {
                    String thing = item.split(" /by ")[0];
                    String time = item.split(" /by ")[1];
                    tasks.add(new Deadline(thing, time));
         } else {
                    ui.reply("Can read instructions or not? Deadline cannot be empty :/");
         }
}

Thus inside this while loop it is just:
while ...
            } else if (task.equals("deadline")) {
                          deadlineTask();
           }
...

src/main/java/task/Event.java
```diff
@@ -0,0 +1,33 @@
+package task;
+
+import java.time.LocalDateTime;
+import java.time.format.DateTimeFormatter;
+import java.time.format.FormatStyle;
+
+public class Event extends Task {
+
+    protected LocalDateTime ldt;
+    public String at;
```
brigittesantoso: Class variables should not be public.
image

src/main/java/task/TaskListTest.java
```diff
@@ -0,0 +1,26 @@
...
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+public class TaskListTest {
+    @Test
+    void addTask() {
+        TaskList tl = new TaskList();
+        assertEquals("Ah sure. I've added this task:\n" +
+                "      [T][ ] sleep\n" +
+                "    Now you have 1 task in the list.", tl.addTask(new Todo("sleep"), 1));
```
brigittesantoso: Spacing on this line should be 12 instead of 8 as it is a wrapped line. Same for below case under method removeTask()

src/main/java/util/Storage.java
```diff
@@ -0,0 +1,84 @@
...
+            String s = sc.nextLine();
+            Task t = parseTextFile(s);
+            tasks.add(t);
+            }
+        }
+
+    private static Task parseTextFile(String s) {
+        Task t;
+        String[] strarr = s.split(" ");
+        if (s.contains("[X]")) {
```
brigittesantoso, line 102: Code readability can be improved for this if-else part.

src/main/java/util/Storage.java
```diff
@@ -0,0 +1,84 @@
...
+import java.io.FileNotFoundException;
+import java.io.IOException;
+import java.util.List;
+import java.io.FileWriter;
+import java.util.Scanner;
+
+
+public class Storage {
+
+    File file;
```
brigittesantoso: Class variable scope should be declared and should be non-public.

src/main/java/task/TaskList.java
```diff
@@ -0,0 +1,120 @@
+package task;
+
+import util.Ui;
+
+import java.util.ArrayList;
+import java.util.List;
+
+public class TaskList {
+    public List<Task> list;
```
brigittesantoso: Class variables should never be declared as public.
image

src/main/java/util/Storage.java
```diff
@@ -0,0 +1,84 @@
...
+    public void loadFile(List<Task> tasks) throws FileNotFoundException {
+        Scanner sc = new Scanner(file);
+        while (sc.hasNext()) {
+            String s = sc.nextLine();
+            Task t = parseTextFile(s);
+            tasks.add(t);
+            }
+        }
+
+    private static Task parseTextFile(String s) {
```
brigittesantoso, line 61: Naming of variables here can be improved instead of "String s" and "Task t"

brigittesantoso (commented):
Good job! Just a few minor changes needed to improve code quality 👍

src/main/java/task/Event.java
```diff
@@ -0,0 +1,35 @@
+
+package task;
+
+import java.time.LocalDateTime;
+import java.time.format.DateTimeFormatter;
+import java.time.format.FormatStyle;
+
+public class Event extends Task {
+
+    public String at;
```
SethCKL: Agree with brigittesantoso, recommend setting access modifier to private

src/main/java/task/TaskList.java
```diff
@@ -0,0 +1,125 @@
+
+package task;
+
+import util.Ui;
+
+import java.util.ArrayList;
+import java.util.List;
+
+public class TaskList {
+    public List<Task> list;
```
SethCKL: same as above and in agreement with brigittesantoso, recommend setting variables access modifier to private always, unless the usecase or specific situation calls for otherwise

src/main/java/Duke.java
```diff
@@ -1,10 +1,167 @@
...
+                break;
+            case "event":
+                if (!item.equals("")) {
+                    String thing = item.split(" /at ")[0];
+                    String time = item.split(" /at ")[1];
+                    tasks.add(new Event(thing, time));
+                } else {
+                    ui.reply("Can read instructions or not? Event cannot be empty :/");
+                }
+
```
SethCKL: would suggest grouping up the switch statements to do away the duplicated code.

case "todo":
case "deadline":
case "event":

if (!item.equals("")) {
Check if it's deadline/event. if so call the additional split functions.
else, just create a todo task
then add task at the end

}

src/main/java/Duke.java
SethCKL: agreed, can potentially abstract out certain parts of the code to another class or to another function

PR #211, opened 2022-01-26
[Kah Leong] iP

kahleongq:
### Duke

- text-based
- easy to learn
- ~~FAST~~ SUPER FAST to use

All you need to do is,

2. double-click it.
3. add your tasks.

And it is **FREE**!

Features:

- [ ]  Managing tasks 
- [ ] Reminders (coming soon)

```
public class Duke {
    }
}
```


src/main/java/Duke.java
```diff
@@ -1,10 +1,124 @@
...
+            + "|  _ \\ _   _| | _____ \n"
+            + "| | | | | | | |/ / _ \\\n"
+            + "| |_| | |_| |   <  __/\n"
+            + "|____/ \\__,_|_|\\_\\___|\n";
         System.out.println("Hello from\n" + logo);
+
+        BufferedReader br = new BufferedReader(new InputStreamReader(System.in));
+        String nextLine = br.readLine().strip().replaceAll(" +", " ");
+
+        ArrayList<Task> list = new ArrayList<Task>();
```
robinrojh: Using a longer variable name (e.g. taskList) would help in case you need to change the calls for this variable!

src/main/java/Duke.java
```diff
@@ -1,10 +1,124 @@
...
+
+        while (!nextLine.equals("bye")) {
+            System.out.println("__________________________________________________");
+            String firstWord = nextLine.split(" ")[0];
+            if (firstWord.equals("mark")) {
+                try {
+                    int indexToMark = Integer.parseInt(nextLine.substring(5)) - 1;
```
robinrojh: In cases like this, it could be cleaner to keep the variable name shorter as the scope of this variable is only in the try block.

src/main/java/Duke.java
```diff
@@ -1,10 +1,124 @@
...
+                            + list.get(indexToMark));
+                } catch (IndexOutOfBoundsException e) {
+                    System.out.println("Please choose a valid task! (Your list has " 
+                            + list.size() + " tasks)");
+                } catch (NumberFormatException e) {
+                    System.out.println("Please input an integer!");
+                }
+            } else if (firstWord.equals("unmark")) {
+                try {
+                    int indexToUnmark = Integer.parseInt(nextLine.substring(7)) - 1;
```
robinrojh: Same here! However, you can decide for yourself. If you feel like you will use this variable many times and in deeper blocks, you can keep it this way.

src/main/java/Duke.java
```diff
@@ -1,10 +1,124 @@
...
-                + "| | | | | | | |/ / _ \\\n"
-                + "| |_| | |_| |   <  __/\n"
-                + "|____/ \\__,_|_|\\_\\___|\n";
+            + "|  _ \\ _   _| | _____ \n"
+            + "| | | | | | | |/ / _ \\\n"
+            + "| |_| | |_| |   <  __/\n"
+            + "|____/ \\__,_|_|\\_\\___|\n";
         System.out.println("Hello from\n" + logo);
+
+        BufferedReader br = new BufferedReader(new InputStreamReader(System.in));
```
robinrojh: Since the BuefferedReader br has a very large scope (accessible everywhere main method), you should most likely name this a bit longer so that it is easier to modify or search for in the future.

robinrojh (commented):
The only part that you seem to miss is the naming of variables-- storing variables with a larger scope with a longer name and storing variables with a smaller scope with a shorter name. Overall, you have met most of the coding standards.

src/main/java/Duke.java
```diff
@@ -1,10 +1,124 @@
...
+                if (nextLine.split(" ").length > 1) {
+                    System.out.println("Uh-oh! There should not be any words after 'list'.");
+                } else if (list.isEmpty()) {
+                    System.out.println("Uh-oh! List is empty!");
+                } else {
+                    for (int i = 0; i < list.size(); i++) {
+                        System.out.println((i + 1) + "." + list.get(i));
+                    }
+                }
+            } else if (firstWord.equals("delete")) {
```
simjunheng: Perhaps it would be good if you can create more classes e.g. Parser with specific purposes. This way, you might be able to avoid deep nesting

src/main/java/Event.java
```diff
@@ -0,0 +1,28 @@
+public class Event extends Task {
```
simjunheng: Maybe it would be a good idea to include class description? This way, fellow developers can catch a glimpse of the class purpose

src/main/java/Event.java
```diff
@@ -0,0 +1,28 @@
...
+        super(description);
+        this.at = at;
+    }
+
+    public Event(String description, String at, boolean isDone) {
+        super(description, isDone);
+        this.at = at;
+    }
+    
+    @Override
```
simjunheng: Perhaps it would be a good idea to include method descriptions

simjunheng (commented):
The absence of java docs and code quality violations may be due to the incompletion of week 3 IP tasks. These problems should naturally be gone once week 3 tasks are completed. You can do it! 👍

src/main/java/Duke.java
```diff
@@ -1,10 +1,198 @@
...
+                if (isMarked) {
+                    task = task.mark();
+                }
+                list.add(task);
+            } else if (String.valueOf(line.charAt(1)).equals("D")) {
+                String tempDescription = line.split("by: ")[0].substring(7);
+                int tempDescLength = tempDescription.length();
+                String description = tempDescription.substring(0, tempDescLength - 2);
+                String tempTimeBy = line.split("by: ")[1];
+                int endIdx = tempTimeBy.lastIndexOf(")");
```
KristopherPTaslim: perhaps it would be better to write endIndex instead of the short form.

src/main/java/Duke.java
```diff
@@ -1,10 +1,198 @@
...
+                if (isMarked) {
+                    task = task.mark();
+                }
+                list.add(task);
+            } else if (String.valueOf(line.charAt(1)).equals("E")) {
+                String tempDescription = line.split("at: ")[0].substring(7);
+                int tempDescLength = tempDescription.length();
+                String description = tempDescription.substring(0, tempDescLength - 2);
+                String tempTimeBy = line.split("at: ")[1];
+                int endIdx = tempTimeBy.lastIndexOf(")");
```
KristopherPTaslim: perhaps it would be better to write endIndex instead of the short form.

src/main/java/Event.java
```diff
@@ -0,0 +1,35 @@
...
+    protected LocalDate at;
+
+    public Event(String description, String at) {
+        super(description);
+        this.at = LocalDate.parse(at);
+    }
+
+    public Event(String description, LocalDate at, boolean isDone) {
+        super(description, isDone);
+        this.at = at;
```
KristopherPTaslim: maybe can choose different variable naming cause 'at' might be ambiguous

src/main/java/Deadline.java
```diff
@@ -0,0 +1,35 @@
+import java.time.LocalDate;
+import java.time.format.DateTimeFormatter;
+
+class Deadline extends Task {
+    protected LocalDate by;
+
+    public Deadline(String description, String by) {
+        super(description);
+        this.by = LocalDate.parse(by);
```
KristopherPTaslim: perhaps can change the variable naming of by

KristopherPTaslim (commented):
Overall, a relatively well variable naming and java coding standard are followed well! Goodluck on the incoming tasks.

PR #221, opened 2022-01-26
[Lai Chok Hoe] iP

lchokhoe:
**DukeLCH**

DukeLCH is a useful aid that helps you remember things you have to do. It's,

- Text-based
- Easy to learn

All you need to do is,

1. download it from [here](https://github.com/lchokhoe/ip)
2. double-click it
3. add your tasks
4. let it manage your tasks for you :wink:

And it is **FREE**!

Features:

- [x]  Managing Tasks
- [ ]  Managing Deadlines
- [ ]  Reminders (coming soon) 

```
public class Duke {
            new Duke("data\\duke.txt").run();
    }
}
```


src/main/java/duke/Commands.java
```diff
@@ -0,0 +1,297 @@
...
+
+    /**
+     * A method that, when called, uses the provided String[] tokens to build a description for a ToDos Task.
+     *
+     * It then gets the inputted instance of DukeHistory to add a ToDos task entry into it's record
+     * using the generated description.
+     *
+     * @param tokens A String[] of tokens inputted by the user.
+     * @param history An instance of DukeHistory.
+     */
```
jiale-c: I like that your javadoc is very clear and to the point :smiley:

src/main/java/duke/Deadlines.java
```diff
@@ -0,0 +1,55 @@
+package duke;
+
+/**
+ * Represents Deadlines class that inherits from Task class.
+ */
+public class Deadlines extends Task {
+    private final String time;
```
jiale-c, line 7: I like that you used non public variables.

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,200 @@
...
+    public void run() {
+        ui.printGreeting();
+        Scanner io = new Scanner(System.in); // Scanner object created
+        label:
+        while (true) {
+            String input = io.nextLine();
+            String[] tokens = input.split("\\s");
+            String keyword = tokens[0];
+            switch (keyword) {
+            case "bye":
```
jiale-c: I like the use of switch here to run through the possibilities.

jiale-c (commented):
Overall code is very neat and the javadoc is very helpful to understand code.

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,200 @@
...
+                }
+            }
+            default:
+                try {
+                    throw new MissingFormatArgumentException("invalid keywords");
+                } catch (MissingFormatArgumentException ex) {
+                    ui.printListOfCommands();
+                    break;
+                }
+            }
```
whoisjunhong: I like how you managed to account for the different commands in `run()` using a `switch` statement, would it be better for `run()` to have a layer of abstraction such as a `Parser` to handle the different cases in the `switch` statement?

src/main/java/duke/DukeHistory.java
```diff
@@ -0,0 +1,266 @@
...
+     */
+    public void addToDo(String description) {
+        ToDos tempToDo = new ToDos(description);
+        records.add(tempToDo);
+        String msg = "_______________________________________________________\n"
+                + "Understood, adding this task now:\n"
+                + "    " + tempToDo.getToDo()
+                + "Currently you have " + records.size() + " tasks in our records.\n"
+                + "_______________________________________________________\n";
+        System.out.println(msg);
```
whoisjunhong: Should this be abstracted into `DukeUi.java`? This applies to all `String msg` in this file. 😄

src/main/java/duke/Commands.java
```diff
@@ -0,0 +1,297 @@
...
+            DateFormat outputFormat = new SimpleDateFormat("hh:mmaa");
+            try {
+                Date dukeTime = inputFormat.parse(splicedTime);
+                return outputFormat.format(dukeTime);
+            } catch (ParseException e) {
+                e.printStackTrace();
+                return time;
+            }
+        }
+    }
```
whoisjunhong: I like how to accounted for many scenarios to convert the date to a DukeDateTime. 👍 

However, should this be in `Commands.java`? Maybe it would be better to have a `Validation` class to simplifying the conversion and validation of a DukeDateTime.

whoisjunhong (commented):
Overall, code integrity and logic looks great! 🎉 

The common issue found is that many UI is not abstracted and would lead to many bugs and formatting issues when making a change in the future. 🤕 

Perhaps a more intuitive way of calling UI methods can be used throughout your Duke project? 😄

src/main/java/Event.java
```diff
@@ -8,16 +8,17 @@ public Event(String description, String date, String time) {
...
+        this.time = time;
     }
 
     public String getEvent() {
         return "[E]" + this.getTask() + "(at: " + this.date + ", " + this.time + ")\n";
     }
 
     public String getFormattedText() {
-        return "E:" + this.getMark() + ":" + this.getDescription() + ":" + this.timeFrame;
+        return "E>" + this.getMark() + ">" + this.getDescription() + ">" + this.date + ">" + this.time;
```
gigituang: I think we can use this.type instead of hardcoding "E" for the event. This can be done for the other tasks too:)

src/main/java/ToDos.java
```diff
@@ -13,6 +13,6 @@ public String getToDo() {
     }
 
     public String getFormattedText() {
-        return "T:" + this.getMark() + ":" + this.getDescription();
+        return "T>" + this.getMark() + ">" + this.getDescription();
```
gigituang: Same here, you can use this .type.

src/main/java/duke/Commands.java
```diff
@@ -270,4 +270,28 @@ public void delete(int index, DukeHistory history)  {
...
+            phrase.append(tokens[i]);
+            if (i != (tokens.length - 1)) {
+                phrase.append(" ");
+            }
+        }
+        String border = "_______________________________________________________\n";
+        System.out.println(border
+                + "Here is what we found:\n"
+                + history.findPhrase(phrase.toString()) + border);
+    }
```
gigituang: This command class may be a little too long?
It would be great if you could split up the find and delete command, and extend them from a command class. It will be easier to sort and debug if you encounter any issues:)
an example could be `public class FindCommand extends Command {}`

src/main/java/duke/Commands.java
```diff
@@ -8,18 +8,34 @@
...
     }
 
+    /**
+     * A method that takes in a String, checks it against 2 accepted date formats
+     * and converts it into a default date format for output. If the String
+     * satisfies neither format, it is simply outputted as is.
+     *
+     * @param date A String input that represents a date.
+     * @return Either a default formatted date or the user input.
+     */
```
gigituang: Really love your detailed comments and how it clearly explains the use of each functions!:)

gigituang (commented):
Overall, I love how detailed your codes are as I can understand it pretty easily. Maybe one thing to note is to keep your classes a little shorter as it is pretty lengthy. You can consider splitting into smaller classes if that makes sense. Really good work! Keep it up!!

PR #223, opened 2022-01-26
[Tan Li En, Ryan] iP

ryantanlien:
# Butler
> "Your mind is for having idea, not holding them." - David Allen

Butler frees your mind of having to remember things you need to do. It's 
- text-based
- easy to learn
- ~~FAST~~ SUPER FAST to use

All you need to do is,

1. download it from [here](https://github.com/ryantanlien/ip)
2. double click it
3. add your tasks
4. let it manage your tasks for you :wink:

And it's **FREE**!

Features:
- [x] Managing tasks
- [x] Managing deadlines
- [ ] Reminders (coming soon)

#

If you're a Java programmer, you can use it to practice Java too. Here's the `main` method

```java
public class Main {
    }
}
```


src/main/java/butler/util/Parser.java
```diff
@@ -0,0 +1,136 @@
...
+import butler.command.ByeCommand;
+import butler.command.DeadlineCommand;
+import butler.command.DeleteCommand;
+import butler.command.EventCommand;
+import butler.command.InvalidCommand;
+import butler.command.ListCommand;
+import butler.command.MarkCommand;
+import butler.command.TodoCommand;
+import butler.command.UnmarkCommand;
+import butler.command.FindCommand;
```
redpelican2108: Shouldn't butler.command.FindCommand be Positioned after EventCommand and InvalidCommand?

src/main/java/butler/task/Event.java
```diff
@@ -0,0 +1,47 @@
+package butler.task;
+
+import butler.util.LocalDateTimeManager;
+
+public class Event extends Task {
+    
+    /* String representation of date and time not following any format */
```
redpelican2108: Shouldn't Javadoc of class members use /** instead of /* ?

src/main/java/butler/command/FindCommand.java
```diff
@@ -0,0 +1,19 @@
...
+import butler.util.Storage;
+import butler.util.Ui;
+
+public class FindCommand extends Command {
+    private String searchString;
+
+    public FindCommand(String searchString) {
+        this.searchString = searchString;
+    }
+
```
redpelican2108, line 13: Isn't Find Command missing it's javadoc?

redpelican2108 (commented):
I like how your code is very clean and easy to read. Keep up the good work

src/main/java/butler/util/ui/DialogBox.java
```diff
@@ -0,0 +1,49 @@
...
+
+    /** Default constructor
+     *
+     * @param l Label component of the DialogBox
+     * @param iv Image associated to the DialogBox, typically that of the chat bot or user
+     */
+    public DialogBox(Label l, ImageView iv) {
+        text = l;
+        displayPicture = iv;
+
```
justinekoh, line 25: Nice use of blank lines to separate logical code blocks! :)

src/main/java/butler/util/Parser.java
```diff
@@ -0,0 +1,137 @@
...
+            if (stringArray.length != 2) {
+                throw new ButlerInputException("Error. Sorry Master, " +
+                        "but a mark or unmark command must be followed by one Integer.\n");
+            }
+            try {
+                return new MarkCommand(Integer.parseInt(stringArray[1]) - 1);
+            } catch (NumberFormatException exception) {
+                throw new ButlerInputException("Error. Sorry Master, " +
+                       "but a mark or unmark command must be followed by one Integer.\n");
+            }
```
justinekoh: Is there a reason why you chose to handle the case where `stringArray.length != 2` explicitly instead of leaving it to the try-catch block?

justinekoh (commented):
LGTM! Very nice structure and consistent formatting. :)

src/main/java/butler/TaskList.java
```diff
@@ -0,0 +1,193 @@
...
+        while (i < tasks.size()) {
+            Task task = tasks.get(i);
+            if (task.getDescription().contains(searchString)) {
+                output += String.valueOf(i + 1) + ". " + task.toString() + "\n";
+            }
+            i++;
+        }
+        return output;
+    }
+
```
joey-chance, line 59: I like how clean your code is. Good job and keep it up! 👍

src/main/java/butler/TaskList.java
```diff
@@ -0,0 +1,193 @@
...
+     *
+     * @param description description of the Task, which is a Deadline.
+     * @param dateTime String representing the date and time to complete the Deadline by.
+     * @return Butler response that confirms success status of adding the Deadline to the TaskList, and reiterates
+     * the task being added and the number of tasks currently in the list.
+     * @throws ButlerInputException Throws a ButlerInputException if user input is not satisfactory.
+     */
+    public String addDeadlineToList(String description, String dateTime) throws ButlerInputException {
+        if (description.equals("")) {
+            throw new ButlerInputException("Error. Sorry Master, but the description of a Deadline cannot be empty.\n");
```
joey-chance, line 94: Any reason why you did it this way instead of handling print statements in your Ui class?

src/main/java/butler/util/ui/DialogBox.java
joey-chance: I agree! Your code is very readable. 👍  Good job and keep it up!

joey-chance (commented):
LGTM. Your code is very readable 👍

src/main/java/butler/TaskList.java
ryantanlien: Hi Joey! I did it this way because of Single Responsibility Principle, having the message in the exception allows me to handle it and print the message should the exception be thrown. Having the Ui handle all the print statements is probably not the best idea if we have to refactor our code (to pass the message to a GUI for example)

src/main/java/butler/util/Parser.java
ryantanlien: Hi Justine! The reason I chose to handle the case of stringArray.length != 2 here is because it is the job of the parser  to check for input. 

In addition, the try catch block should be reserved for sections of code that throws exceptions further down the function call stack. Even if I did place the if statement in the try catch block, the parser would not know what to do with the exception message and would have to rethrow it anyway, so I just placed it outside the try catch block.

PR #224, opened 2022-01-26
[Teoh Jun Jie] iP

junjieteoh:
# Duke

> “Your mind is for having ideas, not holding them.” – David Allen ([source]( https://dansilvestre.com/productivity-quotes))

- text-based
- easy to learn
- ~~FAST~~ _SUPER_  FAST to use 

All you need to do is,
1. download it from [here](https://github.com/junjieteoh/ip).
2. double-click it.
3. add your tasks.
4. let it manage your tasks for you 😜 

And it is **FREE** 👍 !

Features: 
- [x] Managing tasks
- [x] Managing deadlines 
- [ ] Reminders (coming soon)

```
public class Main {
    }
}
```


src/main/java/duke/parser/DateTimeParser.java
```diff
@@ -0,0 +1,35 @@
...
+import java.time.format.DateTimeFormatter;
+import java.time.format.DateTimeParseException;
+import java.time.format.FormatStyle;
+import java.time.LocalDate;
+
+public class DateTimeParser {
+
+
+    /**
+     * Return a LocalDate object based on the input String
```
cpwill01: According to the part of the coding standard on [Javadoc headers](https://se-education.org/guides/conventions/java/intermediate.html#comments), I think this should start with "Returns" rather than "Return". I found similar issues with the header comments in other classes as well.

src/main/java/duke/parser/DateTimeParser.java
```diff
@@ -0,0 +1,35 @@
...
+     *
+     * @param date an String giving the date to be parsed
+     * @return a LocalDate object from the parsed String.
+     * @throws DateTimeParseException
+     */
+
```
cpwill01: Should this empty line be deleted?

src/main/java/duke/parser/Parser.java
```diff
@@ -0,0 +1,214 @@
...
+    public Parser(String userInput) {
+        this.userInput = userInput;
+    }
+
+    /**
+     * Process user commands.
+     *
+     * @return False if the command is "bye", True otherwise
+     */
+    public boolean processInput() {
```
cpwill01: Should this return a boolean? Based on my understanding, I think methods that return a boolean should also follow the `isXYZ()` naming convention (eg. `isExit()`, `isDone()`).

cpwill01: Perhaps you could consider separating this into 2 functions where 1 solely handles displaying the correct response while another handles checking whether the application should exit?

src/main/java/duke/parser/Parser.java
```diff
@@ -0,0 +1,214 @@
...
+            }
+
+            break;
+        default:
+            throw new InvalidTypeException("Invalid type");
+        }
+
+        updateFile();
+
+        String output = "   __________________________________________________\n"
```
cpwill01: Should this line and line 204 below use `Ui.createLine()` instead?

src/main/java/duke/parser/Parser.java
```diff
@@ -0,0 +1,214 @@
...
+        this.userInput = userInput;
+    }
+
+    /**
+     * Process user commands.
+     *
+     * @return False if the command is "bye", True otherwise
+     */
+    public boolean processInput() {
+        if(userInput.equals("bye")) {
```
cpwill01: There should be a space between `if` and the open bracket. I noticed the same issue on line 81 and in some of the other files. Perhaps you could consider using [checkStyle](https://nus-cs2103-ay2122s2.github.io/website/schedule/week4/project.html#4-add-increments-as-branches-a-gradle-a-checkstyle-level-10-a-varargs)? It really helps me a lot in fixing my code to meet the coding standards 😅

src/main/java/duke/parser/Parser.java
```diff
@@ -0,0 +1,214 @@
...
+        } else {
+
+            try {
+                String newTaskMessage = Parser.createTask(userInput);
+                System.out.println(newTaskMessage);
+            } catch (InvalidTypeException e) {
+                String errorMsg = Ui.createLine()
+                        + "       Opps, the command \"" + userInput + "\" is not supported :(\n"
+                        + Ui.createLine();
+                System.out.println(errorMsg);
```
cpwill01: Would there be a way to use the error messages stored inside the exceptions instead of creating a new error message here? Otherwise, is there a need for the exceptions to have a message in their constructors?

cpwill01: Would there be a way to use the error messages stored inside the exceptions instead of creating a new error message here? Otherwise, is there a need for the exceptions to have a message in their constructors?

src/main/java/duke/task/Task.java
```diff
@@ -0,0 +1,146 @@
+package duke.task;
+
+import duke.storage.Storage;
+import java.util.ArrayList;
+
+public class Task {
+    protected String name;
+    protected boolean isDone;
+    protected int id;
+    private static ArrayList<Task> listOfTask = new ArrayList<>(100);
```
cpwill01: Should this be called `tasks` instead of `listOfTask`? [Plural form](https://se-education.org/guides/conventions/java/intermediate.html#naming) should be used for names representing collections.

src/main/java/duke/task/Task.java
```diff
@@ -0,0 +1,146 @@
...
+
+        return result;
+    }
+
+    /**
+     * Retrieve a formatted String of the current task list.
+     *
+     * @return a String with the current task list
+     */
+    public static String printArray() {
```
cpwill01: Should this and the overloaded method below be called `printTaskList()` instead of `printArray()`? I got a little confused when I saw `Task.printArray()` being used in the other classes because I wasn't sure what array it's referring to.

src/main/java/duke/task/Task.java
```diff
@@ -0,0 +1,146 @@
...
+                }
+            }
+
+        }
+
+        return output;
+    }
+
+    /**
+     * Retrieve a formatted String of the current task list.
```
cpwill01: Should this description be changed? I understood the method as retrieving the string of only the part of the task list specified by the input array rather than the full task list.

src/main/java/duke/task/Task.java
```diff
@@ -0,0 +1,146 @@
...
+     *
+     * @param array an integer array containing task number to be printed
+     * @return a String with the current task list
+     */
+    public static String printArray(Integer[] array) {
```
cpwill01: Could the parameter name be more descriptive rather than `array`?

cpwill01 (commented):
Overall, I found your code mostly easy to follow. However, I did spot some coding standard violations. I've pointed out some of them, but I may have missed some out so I really recommend setting up checkStyle as it helped me spot all the small errors in my own code. There were also some code quality issues I pointed out, but most of them should be quick to fix. All the best for the rest of the iP and tP! 😄

src/main/java/duke/parser/DateTimeParser.java
```diff
@@ -0,0 +1,35 @@
+package duke.parser;
+
+import java.time.format.DateTimeFormatter;
+import java.time.format.DateTimeParseException;
+import java.time.format.FormatStyle;
+import java.time.LocalDate;
+
+public class DateTimeParser {
+
```
HanJiyao, line 9: maybe can add missing class JavaDoc here?

src/main/java/duke/storage/Storage.java
```diff
@@ -0,0 +1,57 @@
...
+        try {
+            writeToFile(textToAdd);
+        } catch (IOException e) {
+            System.out.println("Something went wrong: " + e.getMessage());
+        }
+    }
+
+    public static String getCurrentFilePath() {
+        return filePath;
+    }
```
HanJiyao: maybe this method is never used?

src/main/java/duke/task/Task.java
```diff
@@ -0,0 +1,146 @@
...
+import duke.storage.Storage;
+import java.util.ArrayList;
+
+public class Task {
+    protected String name;
+    protected boolean isDone;
+    protected int id;
+    private static ArrayList<Task> listOfTask = new ArrayList<>(100);
+    private static int counter = 0;
+
```
HanJiyao, line 12: Since you define list of task as ArrayList, is it better to check array size directly?

src/main/java/duke/ui/Ui.java
```diff
@@ -0,0 +1,41 @@
...
+        return goodByeMessage;
+    }
+
+    /**
+     * Print divider line.
+     * @return divider line
+     */
+    public static String createLine() {
+        return  "   __________________________________________________\n";
+    }
```
HanJiyao: Should it can represent this line as constant DIVIDER to be more straight forward as it is only one String?

src/main/java/duke/task/Deadline.java
```diff
@@ -0,0 +1,22 @@
...
+        super(name);
+        this.time = time;
+        this.type = "D";
+    }
+
+    @Override
+    public String toString() {
+        String status = this.isDone ? "X" : " ";
+        return "[" + this.type + "][" + status + "]" + this.name + " (by: "
+                + DateTimeParser.formatDate(this.time) + ")";
```
HanJiyao, line 33: according to code standard the breaks should occurs outside the parenthesized expression
longName1 = longName2 * (longName3 + longName4 - longName5)
        + 4 * longname6

HanJiyao (commented):
Overall, I found your code easy to read for the most part. I noted a few minor coding standard violations too. Some of the classes are getting quite long. Consider splitting into
smaller classes like different command if that makes sense.

PR #230, opened 2022-01-26
[Cheng Zi Wei Keith] iP

KeithCZW:
# Welcome to Duke!
> "To eat an apple and expect to have a cake after is foolishness" - Sun Tzu, probably

Duke is a chatbot that will help you manage tasks! It's,
- text-based
- easy to learn
- ~~FAST~~ __GIGA FAST__ to use!

How to use it? All you have to do is,
1. download it from [here](https://github.com/KeithCZW/ip/releases/tag/A-Jar).
2. double-click it.
3. add your tasks
4. tasks = managed 😄

It's also **FREE!!!!**

Features include:
- [x] Managing tasks
- [ ] Reminders (coming soon)

---

```java
public class Main {
    }
}
```


src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,103 @@
+package duke;
+
+import duke.task.Deadline;
+import duke.task.Event;
+import duke.task.Task;
+import duke.task.Todo;
+
+import java.io.*;
```
iztanpy: Might not be accurate to import java generic, maybe you could import the specific classes utilised instead?

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,151 @@
...
+     * Parses the user input and splits it up for the Todo object to read
+     *
+     * @param input User input
+     * @return Todo description
+     * @throws DukeException If user enters invalid input, throws an exception
+     */
+    public String parseTodo(String input) throws DukeException {
+        String[] split = input.split(" ");
+        if (split.length < 2) {
+            throw new DukeException("OH NO! The description of event cannot be empty.");
```
iztanpy: I did something similar but an alternative could be

throw new DukeException(e.getMessage());

iztanpy (commented):
Nicely done! Just suggested some small alternatives and coding standard fixes 👍

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,302 @@
...
+                    break;
+                case FIND:
+                    String word = parser.parseFind(input);
+                    ui.listTasks(tasks.findTasks(word));
+                    break;
+                case ERROR:
+                    throw new DukeException("Sorry :( I don't know what this means.");
+                }
+            } catch (DukeException e) {
+                ui.sayMessage(e.getErrorMsg());
```
lamwj98: would it be better to handle this cases somewhere else or create another class to avoid populating the duke class?

src/main/java/duke/DukeException.java
```diff
@@ -0,0 +1,26 @@
...
+    }
+
+    /**
+     * Gets the error message assigned to DukeException
+     *
+     * @return The error message
+     */
+    public String getErrorMsg() {
+        return errorMsg;
+    }
```
lamwj98, line 25: what is the purpose of creating a getErrorMsg function when there is a .getMessage() function from exception?

src/main/java/duke/task/Deadline.java
```diff
@@ -0,0 +1,94 @@
...
+            throw new DukeException("Invalid date");
+        }
+    }
+
+    /**
+     * Getter for time/date of deadline
+     *
+     * @return time/date
+     */
+    public String getBy() {
```
lamwj98, line 68: naming of variable is not easily understand, maybe getDue would be a better name ?

src/main/java/duke/task/Event.java
```diff
@@ -0,0 +1,94 @@
...
+import java.time.LocalDate;
+import java.time.LocalTime;
+import java.time.format.DateTimeFormatter;
+import java.time.format.DateTimeParseException;
+
+/**
+ * Event task that inherits from Task
+ */
+public class Event extends Task {
+    protected LocalDate at;
```
lamwj98, line 14: same issue, naming of variable not easily understandable, would `date` be a better name?

src/main/java/duke/task/Event.java
```diff
@@ -0,0 +1,94 @@
...
+            throw new DukeException("Invalid date/time");
+        }
+    }
+
+    /**
+     * Getter for time/date of event
+     *
+     * @return time/date
+     */
+    public String getAt() {
```
lamwj98, line 68: same as above, would getEventDate or getDate be better?

src/main/java/duke/task/Task.java
```diff
@@ -0,0 +1,84 @@
...
+
+    /**
+     * Sets isDone to false
+     */
+    public void markAsNotDone() {
+        this.isDone = false;
+    }
+
+    /**
+     * Overloaded toString function
```
lamwj98, line 66: Should be overwrite instead of overload

src/main/java/duke/task/Task.java
```diff
@@ -0,0 +1,84 @@
...
+
+    /**
+     * Empty getFileString function for child classes
+     *
+     * @return Nothing
+     */
+    public String getFileString() {
+        return "";
+    }
+}
```
lamwj98, line 84: would it be better to declare as an abstract method?

lamwj98 (commented):
Overall code is easy to read and understandable. Some improvements needed on variables namings to make methods more understandable.

src/main/java/duke/task/Task.java
KeithCZW: Yes, fair enough. I will change this!

src/main/java/duke/task/Deadline.java
KeithCZW: Yeah perhaps some of the naming is too vague, I will change that in the next push.

src/main/java/duke/Duke.java
KeithCZW: Yeah I will probably encapsulate the functionality in another class soon.

src/main/java/duke/Storage.java
KeithCZW: Oops! I missed that. Will fix it.

PR #235, opened 2022-01-26
[Tan De Shao] iP

tandeshao:
# Duke #
> "Why am i doing this, why does the code not work." - Some sad single student out in soc 

**Features include:**
- level-1
- level-2
- ~~level-3~~ **Level-4 NEW AND IMPROVED!**

**All you need to do is**
1. download it from [here](https://www.youtube.com/watch?v=jl9xKw_xEas&ab_channel=TopTunes).
2. Double click it.
3. add your tasks
4. let it manage your tasks for you 😃 

and it is **FREE**!

blahblah:
- [x] black sheep ?
- [x] and the hoo?
- [x] 3 black wolves?
- [x] chicken rice?
***
if you programmer, you can use it to practice java too. Here's the ``main`` method:
```java
public class Main {
    }
}
```


src/main/java/DukeComponent/Ui.java
```diff
@@ -0,0 +1,58 @@
...
+     * Runs the Ui for the Duke program.
+     * @param sc Scanner to allow continuous user input from the CLI.
+     * @param tasks TaskList that is manipulated by the {@link DukeComponent.Command} class.
+     * @param storage Storage that is manipulated by the {@link DukeComponent.Command} class.
+     */
+    public void run(Scanner sc, TaskList tasks, Storage storage) {
+        while (sc.hasNext()) {
+            String userInput = sc.nextLine();
+            try {
+
```
kerbysoh: There's an unnecessary blank line here.

src/main/java/DukeComponent/Command.java
```diff
@@ -0,0 +1,131 @@
...
+     * Execution of command logic in Duke.
+     */
+    public void execute() {
+        String[] wordSplit = userInput.split(" ");
+        String action = wordSplit[0];
+        String[] split = userInput.split("/");
+        int start = userInput.indexOf(" ") + 1;
+        int end = userInput.lastIndexOf('/');
+        String task = end == -1 ? "" : userInput.substring(start, end - 1);
+        String details = split.length > 1 ? split[1].substring(3) : "";
```
kerbysoh: Avoid the use of magic numbers here like 1 and 3

src/main/java/DukeComponent/Command.java
```diff
@@ -0,0 +1,131 @@
...
+
+        if (userInput.equals("bye")) {
+            sayBye();
+            return;
+        }
+
+        switch (action) {
```
kerbysoh: You can make use of the default branch here to handle your commands that are invalid

src/main/java/Tasks/Tasks.java
```diff
@@ -0,0 +1,79 @@
...
+import java.time.format.DateTimeParseException;
+
+/**
+ * A class that belongs to the Tasks Package.
+ * This class encapsulates the logic of how Tasks should be represented in Duke.
+ */
+abstract public class Tasks {
+
+    private String task;
+    private Boolean marked;
```
kerbysoh: Boolean here should be named isMarked rather than marked

src/main/java/Tasks/Tasks.java
```diff
@@ -0,0 +1,79 @@
...
+     * @return
+     */
+    abstract public String cacheString();
+
+    /**
+     * Converts a date of format "yyyy-MM-dd" to "MMM dd yyyy".
+     * @param s String representation of a date.
+     * @return Date that follows the format if applicable.
+     */
+    public static String returnDate(String s) {
```
kerbysoh: Function name could be better named to describe the function (To convert date from one form to another). May I suggest using convertDate instead?

kerbysoh (commented):
Well done! Code is very well documented, clean and well written. There are just a few pointers I've noticed in my review that you may want to look into. :)

src/main/java/Duke.java
```diff
@@ -1,10 +1,52 @@
...
+    private final Storage storage;
+    /**
+     * {@link TaskList}
+     */
+    private final TaskList tasks;
+
+    /**
+     * {@link Ui}
+     */
+    private final Ui ui;
```
yusufaine: solid use of `@link` 👍

src/main/java/DukeComponent/Command.java
```diff
@@ -0,0 +1,131 @@
...
+    /**
+     * Deletes the task in the TaskList.
+     * @param i 0-based index for deletion of Task in the TaskList.
+     */
+    private void deleteTask(int i) {
+        System.out.println("Noted. I've removed this task:");
+        System.out.println(tasks.remove(i));
+        String s = String.format("Now you have %d tasks in the list.", tasks.size());
+        System.out.println(s);
+    }
```
yusufaine: You can consider moving all your print statements into `UI` so that your code is less cluttered and easier to focus on what the logic is (even though it's quite short here, but still)

src/main/java/DukeComponent/Parser.java
```diff
@@ -0,0 +1,78 @@
...
+     * Helper method for {@link #checkUserInput(String)}.
+     * @param command Checks if user input is a command.
+     * @return True if input is not a command and false otherwise.
+     */
+    private static boolean notCommand(String command) {
+        return (!command.equals("bye") && !command.equals("list")
+                && !command.equals("delete") && !command.equals("mark") && !command.equals("unmark")
+                && !command.equals("todo") && !command.equals("deadline") && !command.equals("event") &&
+                !command.equals("find"));
+    }
```
yusufaine: This can be replaced by adding a default case as mentioned by @\kerbysoh in `Command`

src/main/java/DukeComponent/TaskList.java
```diff
@@ -0,0 +1,98 @@
...
+    public int size() {
+        return ls.size();
+    }
+
+    /**
+     * Prints out a list of Task in that TaskList that matches the description.
+     * @param description  Description of the Task.
+     */
+    public void find(String description) {
+        int count = 1;
```
yusufaine: I think this variable a bit redundant since you're not using/returning it.

```suggestion

```

src/main/java/DukeComponent/TaskList.java
```diff
@@ -0,0 +1,98 @@
...
+        for (Tasks t : ls) {
+            if (t.getTask().contains(description)) {
+                System.out.println(count + "." + t);
+                count++;
```
yusufaine:
```suggestion

```

src/main/java/DukeComponent/TaskList.java
```diff
@@ -0,0 +1,98 @@
...
+                return ls.get(currentIndex++);
+            }
+
+            @Override
+            public void remove() {
+                throw new UnsupportedOperationException();
+            }
+        };
+
+    }
```
yusufaine: I did this too for mine initially but then I realised that you don't actually have to implement an iterator and create a getter for your taskList:

```suggestion
    public ArrayList<Tasks> getList() {
        return this.ls;
    }
```

that way when you want to do a for-each, you can call:
* `taskList.getList().forEach(...)` or 
* `for (Task t : taskList.getList())`

src/main/java/DukeComponent/Ui.java
```diff
@@ -0,0 +1,58 @@
+package DukeComponent;
```
yusufaine: Mentioned earlier but you can consider moving all your printline stuff here since this class is meant to interact with the user.

src/main/java/Tasks/Tasks.java
```diff
@@ -0,0 +1,79 @@
...
+
+import java.time.LocalDate;
+import java.time.format.DateTimeFormatter;
+import java.time.format.DateTimeParseException;
+
+/**
+ * A class that belongs to the Tasks Package.
+ * This class encapsulates the logic of how Tasks should be represented in Duke.
+ */
+abstract public class Tasks {
```
yusufaine: nit-picky: I think this class should be named `Task` instead

src/main/java/Tasks/DeadLines.java
```diff
@@ -0,0 +1,43 @@
...
+     * @return String representation of DeadLine.
+     */
+    @Override
+    public String toString() {
+        if (this.getMarked()) {
+            return "[D]" + "[X" + "] " + this.getTask() + " (by: " + returnDate(this.date) + ")";
+        } else {
+            return "[D]" + "[ " + "] " + this.getTask() + " (by: " + returnDate(this.date) + ")";
+        }
+    }
```
yusufaine: You can consider using `String.format()` to make this a little more readable, here's an example of mine (although slightly different use-case):

```java
public String toFile() {
    String markStatus = super.isMarked ? "1" : "0";
    SimpleDateFormat sdf = new SimpleDateFormat("dd/MM/yyyy HH:mm");
    return String.format("%s | %s | %s | %s\n", "D", markStatus, super.taskName, sdf.format(this.by));
}
```

yusufaine (commented):
Overall clean code, not much to comment about apart from some nit-picky pointers.

src/main/java/component/TaskList.java
```diff
@@ -0,0 +1,100 @@
...
+            @Override
+            public Tasks next() {
+                return ls.get(currentIndex++);
+            }
+
+            @Override
+            public void remove() {
+                throw new UnsupportedOperationException();
+            }
+        };
```
fantablack: I like this use of an Iterator to iterate through the tasks compared to the very common for/foreach loops! 👍

src/main/java/exceptions/DeadlineException.java
```diff
@@ -0,0 +1,25 @@
+package exceptions;
+
+/**
+ * A class that belongs to the Exceptions Package.
+ * This class encapsulates the message that should be displayed when users parsed an invalid input
+ * for the {@link tasks.DeadLines} class.
+ */
+public class DeadlineException extends TaskException {
```
fantablack, line 8: I like that you used inheritance to differentiate the different types of exceptions, instead of simply using different Strings in the respective exceptions' messages

fantablack (commented):
Overall, I liked that your code is clean and easy to follow. Interesting use of iterator to process your Lists. Keep up the good work!

PR #249, opened 2022-01-27
[Kaaviya Selvam] iP

kavxya:
# HelperBot  
>“Your mind is for having ideas, not holding them.” – David Allen [source](https://dansilvestre.com/productivity-quotes)  

DukePro frees your mind of having to remember things you need to do. 
It's,  
-text-based 
-easy to learn 
-~~_FAST_~~ SUPER FAST to use  

All you need to do is,  
1.download it
2.double-click it. 
3.add your tasks. 
4.let it manage your tasks for you 😉  

And it is **FREE**!  
Features:   
-[x]Managing tasks  
-[ ]Managing deadlines (coming soon)  
-[ ]Reminders (coming soon)


src/main/java/Ui.java
```diff
@@ -0,0 +1,95 @@
...
+    public void emptyInput() {
+        System.out.println(" ☹ OOPS!!! The description of a todo cannot be empty.");
+    }
+
+    /**
+     * Prints the error message for invalid commands.
+     */
+    public void doNotUnderstand() {
+        System.out.println("☹ OOPS!!! I'm sorry, but I don't know what that means :-(");
+    }
```
sheimoria: Should this method name be a verb instead?
```suggestion
    public void displayDoNotUnderstandError() {
        System.out.println("☹ OOPS!!! I'm sorry, but I don't know what that means :-(");
    }
```

src/main/java/Ui.java
```diff
@@ -0,0 +1,95 @@
...
+
+    /**
+     * Prints the message for adding a task to the taskList.
+     */
+    public void addTask(TaskList tasks) {
+        System.out.println("Got it. I've added this task: ");
+        System.out.println(tasks.get(tasks.size() - 1));
+        System.out.println(String.format("Now you have %d tasks in the list.", tasks.size()));
+
+    }
```
sheimoria: Might the method name be misleading? My first impression was that this method adds the task to the task list, but instead, it prints the task added.
```suggestion
    public void displayTaskAdded(TaskList tasks) {
        System.out.println("Got it. I've added this task: ");
        System.out.println(tasks.get(tasks.size() - 1));
        System.out.println(String.format("Now you have %d tasks in the list.", tasks.size()));

    }
```

src/main/java/Task.java
```diff
@@ -0,0 +1,51 @@
...
+     * Changes the value of isDone to be false.
+     */
+    public void unmarkAsDone() {
+        this.isDone = false;
+    }
+
+    /**
+     * To display the task to the user and data log.
+     * @return prints the task in the specified format
+     */
```
sheimoria: I think there should be an empty line between the description and parameter section as per the [coding standard](https://se-education.org/guides/conventions/java/intermediate.html#comments)

src/main/java/Deadline.java
```diff
@@ -0,0 +1,23 @@
...
+        super(description);
+    }
+
+    /**
+     * Returns a formatted string representing the Deadline task.
+     * @return String
+     */
+    public String toString() {
+        return String.format("[D][%s] %s", getStatusIcon(), description);
+    }
```
sheimoria: Maybe could consider adding the `@Override` annotation above all the `toString` methods? May not be necessary (coding standard does not talk about it) but I would think that it is recommended.
```suggestion
    @Override
    public String toString() {
        return String.format("[D][%s] %s", getStatusIcon(), description);
    }
```

sheimoria (commented):
LGTM, almost! Just a few nits to fix 👍

src/main/java/Ui.java
```diff
@@ -0,0 +1,95 @@
...
+     */
+    public void markDone(Task task) {
+        System.out.println("Nice! I've marked this task as done: ");
+        System.out.println(task);
+    }
+
+    /**
+     * Prints the message for marking a task as done.
+     *  @param task is the task which is marked as undone.
+     */
```
aachia:
```suggestion
    /**
     * Prints the message for unmarking a task.
     *  @param task is the task which is unmarked.
     */
```

I think you might have a careless mistake here as the header does not describe correctly what the method does. In this case it is unmarking the task instead of marking.

src/main/java/Task.java
```diff
@@ -0,0 +1,51 @@
...
+     * Changes the value of isDone to be true.
+     */
+    public void markAsDone() {
+        this.isDone = true;
+    }
+
+    /**
+     * Changes the value of isDone to be false.
+     */
+    public void unmarkAsDone() {
```
aachia:
```suggestion
    public void unmarkTask() {
```
Could this method name be confusing? From the original name, I would think that it unmarks a task and shows it as done which is conflicting.

PR #250, opened 2022-01-27
[Si Binh] iP

ghost:
## DukePro

> “Your mind is for having ideas, not holding them.” – David Allen [](https://dansilvestre.com/productivity-quotes)

- text-based
- easy to learn
- ~~FAST~~ SUPER FAST to use

All you need to do is,
1. download it from [here](https://github.com/sibinhho99-nus/ip).
2. double-click it.
3. add your tasks.

And it is **FREE**!

Features:
- [ ] Managing tasks
- [ ] Managing deadlines
- [ ] Managing events

```
public class Main {
    }
}
```


src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,133 @@
...
+                    if (command.length() == 4) {
+                        throw new InputMismatchException("The description of a todo cannot be empty.");
+                    }
+
+
+                    pastTasks.add(parser.parseTodos(command));
+                    storage.save();
+
+                    pastTasks.add(parser.parseTodos(command));
+                    storage.save();
```
lzf834: Good job abstracting these commands away into another class

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,133 @@
...
+                    pastTasks.add(parser.parseEvent(command));
+                    storage.save();
+
+                    ui.printMessage(String.format("""
+                            Got it. I've added this task:
+                            %s
+                            Now you have %d %s in the list.""",
+                            pastTasks.get(pastTasks.size() - 1),
+                            pastTasks.size(),
+                            pastTasks.size() == 1 ? "task" : "tasks"
```
lzf834: Perhaps this string could be abstracted out into a method if you are using it in the same format for the 3 different commands

lzf834 (commented):
Overall, I think the code was very well organized and I had an easy time reading through and understanding. Keep it up!

src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,72 @@
+package duke;
+
+import java.io.*;
```
Btaykb: Might want to explicitly list out the packages imported from java.io instead

src/main/java/duke/Ui.java
```diff
@@ -0,0 +1,27 @@
...
+     * Prints a String and then terminate the line.
+     */
+    public void println(String message) {
+        System.out.println(message);
+    }
+
+    /**
+     * Prints the formatted version of given message to console
+     */
+    public void printMessage(String message) {
```
Btaykb: Maybe could add methods that print specific messages instead of passing the whole message to the method ?
eg. have a method named printWelcome() instead of printMessage("Hi I'm duke")

Btaykb (commented):
Code overall looks neat and good! Methods are easy to understand

src/main/java/duke/Ui.java
```diff
@@ -0,0 +1,27 @@
...
+        System.out.printf("""
+                        ____________________________________________________________
+                        %s
+                        ____________________________________________________________%n""",
```
zhongfu: note: this is not valid syntax in Java versions below 13! text blocks were only [made standard in Java 15](http://openjdk.java.net/jeps/378), and we're [supposed to be using Java 11](https://nus-cs2103-ay2122s2.github.io/website/admin/programmingLanguages.html).

this is pretty important, since -- in the words of the module website -- "[a]ny incompatibilities will be considered as bugs".

instead, you may want to consider using concatenation or something of the sort

src/test/ParserTest.java
```diff
@@ -0,0 +1,43 @@
...
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+public class ParserTest {
+    Parser parser = new Parser();
+
+    @Test
+    public void testGettingType() {
+        assertEquals("bye", parser.getType("bye"),
+                "Type of bye command should be `deadline`");
```
zhongfu, line 12: nice, i like that you've added descriptions for each assert here

src/main/java/duke/TaskToDos.java
```diff
@@ -0,0 +1,23 @@
+package duke;
+
+/**
+ * Represents task of todo type
+ *
+ * @author sibinhho99-nus
+ */
+public class TaskToDos extends Task {
```
zhongfu: each instance of this class refers to a single to-do item instead of multiple to-dos; perhaps `TaskToDo` would be a more appropriate name?

zhongfu: (same for `TaskDeadlines` and `TaskEvents`)

src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,72 @@
...
+        bw.write(this.toString());
+        bw.newLine();
+        bw.close();
+    }
+
+    /**
+     * Returns String representation of the duke.TaskList
+     */
+    @Override
+    public String toString() {
```
zhongfu, line 77: it seems like the logic here is exactly the same as in `TaskList.toString()` -- perhaps you could just return `taskList.toString()`?

(also, feels a little weird to want a String representation of instances of this class, but I digress)

src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,72 @@
...
+                boolean isDone = line.charAt(6) == 'X';
+
+                String name = line.split(" \\(at: ")[0].substring(9);
+
+                String startAt;
+                startAt = line.split(" \\(at: ")[1];
+                startAt = startAt.substring(0, startAt.length() - 1); // removes end bracket
+
+                taskList.add(new TaskEvents(isDone, name, startAt));
+            } else { // T
```
zhongfu, line 55: you might want to have an explicit check that `line.charAt(3)` is indeed `== T`, so that you can throw an error if you find an unexpected type of Task (e.g. from a newer version of your program, or because of users making invalid edits to `storage.txt`, etc)

src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,72 @@
...
+    /**
+     * Loads the task list from storage
+     */
+    public void load() throws IOException {
+        FileReader fr = new FileReader("storage.txt");
+        BufferedReader bufferedReader = new BufferedReader(fr);
+        while (bufferedReader.ready()) {
+            String line = bufferedReader.readLine();
+
+            if (line.charAt(3) == 'D') {
```
zhongfu: what happens when you save a `TaskList` with more than 9 tasks, then try to load it? will this still work as expected?

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,63 @@
...
+
+    /**
+     * Returns event Task object from command
+     */
+    public Task parseEvent(String command) {
+        // Extracts name and time to be done by from event command,
+        // eg. event project meeting /at Mon 2-4pm
+        String temp = command.substring(6);
+        String name = temp.split("/")[0].trim();
+        String startAt = temp.split("/")[1].substring(3).trim();
```
zhongfu, line 51: don't know if it's a concern for you, but this won't catch invalid argument keys (e.g. `/event_date <date>` or `/aaa<date>`)

zhongfu: or command strings like `event`:
- `temp = ""`
- `temp.split("/") = [""]`
- `temp.split("/")[1]` = ?

zhongfu (commented):
your changes were pretty easy to understand and not needlessly verbose, though you may want to consider how things can go wrong when parsing things like user input, especially if you're getting substrings using absolute string indexes -- in some cases, those assumptions may not hold true (e.g. if you have more than 9 tasks to be loaded)

also, you may want to hide your instance fields (e.g. `doneBy`, `startAt`, `taskList`) to avoid leaking implementation details from your classes:
- for instance, `Duke` doesn't have to know how `Task` keeps track of whether a task is done or not -- be it through an Enum that determines the state of a Task, or a boolean, or... whatever.
- you may want to consider using accessor methods here instead, so you won't have to hunt around for usages if you do decide to modify your implementation of Task.

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,63 @@
+package duke;
+
+/**
+ * Deals with making sense of the user command
```
lovemathboy:
```suggestion
 * Deals with making sense of the user command.
```
Shouldn't JavaDoc summaries end with a "."?

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,63 @@
...
+ *
+ * @author sibinhho99-nus
+ */
+public class Parser {
+    /**
+     * Returns index to delete from delete command
```
lovemathboy:
```suggestion
     * Returns index to delete from delete command.
     * @param Delete command.
     * @return Index of the delete command.
```
Shouldn't JavaDoc method summaries include parameters and return statements?

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,63 @@
...
+        return command.split(" ")[0];
+    }
+
+    /**
+     * Returns event Task object from command
+     */
+    public Task parseEvent(String command) {
+        // Extracts name and time to be done by from event command,
+        // eg. event project meeting /at Mon 2-4pm
+        String temp = command.substring(6);
```
lovemathboy, line 49: Doesn't hardcoding a '6' (and '9' later on in parseDeadlines) make the program quite hard to understand? Perhaps try using `Pattern` and `Matcher` from `java.util.regex` to extract out the date?

src/main/java/duke/Storage.java
lovemathboy: I agree with Btaykb, shouldn't all packages imported from java.io be written out to follow the coding standards?

src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,72 @@
...
+
+                String startAt;
+                startAt = line.split(" \\(at: ")[1];
+                startAt = startAt.substring(0, startAt.length() - 1); // removes end bracket
+
+                taskList.add(new TaskEvents(isDone, name, startAt));
+            } else { // T
+                boolean isDone = line.charAt(6) == 'X';
+                taskList.add(new TaskToDos(isDone, line.substring(9)));
+            }
```
lovemathboy, line 58: Instead of using magic numbers, perhaps "abstract" out the 3, 6, and 9 outside as static variables to make the program easier to understand?

lovemathboy (commented):
Overall, I like how you abstracted out the various tasks and created a new class to handle a list, instead of doing it in the main class. However, perhaps it'll be a good idea to fix all your JavaDoc comments to follow the coding standards?

PR #252, opened 2022-01-27
[Chong Kok Leong] iP

SethCKL:
### Cleese

> and how may i help you sir? 😃 

~~not boujee enough~~ want to own a _real butler_? 
This is the next **best** thing for you I guess It's

- text-based
- easy peasy lemon squeezy
- FREE

All you gotta do it

1. download [this](https://github.com/SethCKL/ip/releases/tag/A-Jar)
2. double click to run it
3. talk to you new butler
4. let him manage your life

Features:

- [x] Managing tasks
- [ ] GUI (Coming Soon)

Sample Code:
```java
public class Main {
    }
}
```


src/main/java/Cleese/Cleese.java
```diff
@@ -0,0 +1,43 @@
...
+        String response = "default";
+        while (!response.equals("bye")) {
+            try {
+                response = parser.handleCommand(scanner.nextLine(), taskList, ui, storage);
+            } catch(NoDescException error) {
+            System.out.println("☹ OOPS!!! The description of a todo cannot be empty.");
+            ui.printLine();
+            } catch(Exception error) {
+            System.out.println("☹ OOPS!!! I'm sorry but I don't know what that means :-(");
+            ui.printLine();
```
LapisRaider: Should have indentations here.

src/main/java/Cleese/Parser.java
```diff
@@ -0,0 +1,89 @@
...
+        DateTimeFormatter dateTimeFormatter = DateTimeFormatter.ofPattern("d/MM/yyyy HHmm");
+        // split off first word from the rest of the input
+        String[] input = string.split(" ", 2);
+        String command = input[0];
+        String parameters = null;
+        if (input.length > 1) {
+            parameters = input[1];
+        }
+        switch (command) {
+        case "bye":
```
LapisRaider: Might want to have a linebreak in between the if statement closing bracket and the switch statement to improve readability.

src/main/java/Cleese/Parser.java
```diff
@@ -0,0 +1,89 @@
...
+                String localDateTimeString = split[1].split(" ", 2)[1];
+                LocalDateTime localDateTime = LocalDateTime.parse(localDateTimeString, dateTimeFormatter);
+                if (command.equals("event")) {
+                    newTask = new Event(split[0], localDateTime);
+                }
+                if (command.equals("deadline")) {
+                    newTask = new Deadline(split[0], localDateTime);
+                }
+            }
+            taskList.add(newTask);
```
LapisRaider: Might want to put some linebreaks here to improve readability.

src/main/java/Cleese/TaskList.java
```diff
@@ -0,0 +1,35 @@
...
+    }
+    public void printTaskList() {
+        for (int i = 0; i < taskList.size(); i++) {
+            System.out.print(String.format("%d. %s \n", i+1, taskList.get(i).toString()));
+        }
+    }
+    public boolean removeFromTaskList(Task selectedTaskToDelete) {
+        return taskList.remove(selectedTaskToDelete);
+    }
+    public Task get(Integer i) {
```
LapisRaider: Might want to rename the function instead to be more descriptive.
```suggestion
    public Task getTask(Integer i) {
```

src/main/java/Cleese/TaskList.java
```diff
@@ -0,0 +1,35 @@
...
+        return taskList.get(i);
+    }
+    public void add(Task task) {
```
LapisRaider: Might want to rename the function to be more descriptive.

```suggestion
    public void addTask(Task task) {
```

src/main/java/Cleese/TaskList.java
```diff
@@ -0,0 +1,35 @@
...
+        taskList.add(task);
+    }
+    public Integer size() {
```
LapisRaider: Might want to rename the function to be more descriptive.
```suggestion
    public Integer getSize() {
```

src/main/java/Cleese/TaskList.java
```diff
@@ -0,0 +1,35 @@
...
+        return taskList.size();
+    }
+    public void find(String keyword) {
+        for (int i = 0; i < taskList.size(); i++) {
+            if (taskList.get(i).toString().contains(keyword)) {
+                System.out.print(String.format("%d. %s \n", i+1, taskList.get(i).toString()));
+            }
+        }
+    }
```
LapisRaider: Might want to have linebreaks in between functions to improve readability.

src/main/java/Cleese/Task.java
```diff
@@ -0,0 +1,32 @@
+package Cleese;
+
+public class Task {
+    protected String description;
+    protected boolean isDone;
```
LapisRaider: Might want to change this to private instead of protected. For java, unlike C++ where protected makes the variable inaccessible to other classes besides the derived classes, in java, as long as a class is in the same package, they can access the protected variable.

If that is not your intention, you may want to change to private.

src/main/java/Cleese/Storage.java
```diff
@@ -0,0 +1,50 @@
...
+import java.util.Scanner;
+
+public class Storage {
+    private static String filePath;
+    private static File file;
+
+    public Storage(String filePath) {
+        this.filePath = filePath;
+        this.file = new File(filePath);
+    }
```
LapisRaider: The usage of `this.` keyword. It is recommended to use the `this` keyword only when a field is shadowed by a method or constructor parameter.

image

LapisRaider (commented):
Looks great! Just a few nits to fix.

TLDR:
- `this` keyword is recommended to be used only when a field is shadowed by a method or constructor parameter.
- Protected might want to change to private. Protected keyword in Java makes the variable/function accessible for any class in the same package.
- Missing some indentations.
- Might want to add more linebreaks to improve readability.
- Some suggestions on function names to make them more descriptive.

src/main/java/Cleese/Parser.java
```diff
@@ -0,0 +1,89 @@
...
+            if (command.equals("unmark")) {
+                selectedTask.markAsUndone();
+                ui.printUnmarkedMessage(selectedTask);
+            }
+            ui.printLine();
+            return "mark/unmark";
+        case "delete":
+            int deleteSelectedTaskNum = Integer.parseInt(parameters);
+            System.out.println(deleteSelectedTaskNum);
+            Task deleteSelectedTask = taskList.get(deleteSelectedTaskNum-1);
```
likeabowx: Variable names should preferably be a noun for clarity, for example taskSelectedToDelete. In this case, deleteSelectedTask sounds like a verb

src/main/java/Cleese/Parser.java
likeabowx: I agree with @LapisRaider, in general it is easier to read code with break lines between different sections

src/main/java/Cleese/Task.java
```diff
@@ -0,0 +1,32 @@
+package Cleese;
+
+public class Task {
```
likeabowx: Since this class is never used as an object itself, can consider making this an abstract class instead and have the other classes inherit from it. In that case, the ```toStorageString()``` method can be an abstract method and there will be no need for a ```"placeholder"``` in it.

src/main/java/Cleese/TaskList.java
```diff
@@ -0,0 +1,35 @@
+package Cleese;
+
+import java.util.ArrayList;
+
+public class TaskList {
+    private final ArrayList<Task> taskList;
```
likeabowx: The name of the field can be called tasks instead as it can be confused with the class name and be prone to mistakes referring to the class and object.

likeabowx (commented):
LGTM. No major problem, just some minor suggestions for improvement :slightly_smiling_face:

PR #254, opened 2022-01-27
[Ng Wen Hao Dennis] iP

Denniszedead:
# BIG TECH
> How may I help? 😄 
## About
Big Tech is a revolutionary task list for a corporate leader. It:
- Creates the following tasks, TODO, Deadlines and Event.
- Shows the tasks in the list.
- Deletes the tasks in the list.
- Marks and unmarks the tasks.
## Why use?
- IT'S **FREE** (Duh, do i have to emphasise it further!)
- IT'S ~~Fast~~ **SUPER FAST**
## How to install
1. First, download the .jar [File]()
2. The from the directory where you store the .jar file, enter the following command in the terminal:
     `java -jar ip.jar`
3. Then, BIG TECH will start.
## Have you done all of them ?
- [ ] Download the .jar file?
## How to use
BIG TECH currently supports the following command:
- `` list ``: Shows all the tasks in the task list.
- `` todo <Name of the task>``: Creates a todo task.
## In case you are wondering
The main file is located here:
```java
        // Creates a new duke application and tells it to run.
}
```


src/main/java/duke/dukeexceptions/DukeExceptions.java
```diff
@@ -0,0 +1,11 @@
+package duke.dukeexceptions;
```
hieunm1821, line 1: I suggest the naming of the package is duke.exceptions

src/main/java/duke/dukeexceptions/EmptyDate.java
```diff
@@ -0,0 +1,21 @@
+package duke.dukeexceptions;
+
+/**
+ * The exception for when no date is entered for commands that requires dates.
+ */
+public final class EmptyDate extends DukeExceptions {
```
hieunm1821: Should the name of this exception be EmptyDateException ? :thinking:

src/main/java/duke/storage/Storage.java
```diff
@@ -0,0 +1,80 @@
...
+     */
+    public TaskList getData() throws FileNotFoundException {
+        Scanner fileInput = new Scanner(file);
+        TaskList taskList = new TaskList();
+
+        // Gets the data from the file and creates new task into the task list.
+        while (fileInput.hasNextLine()) {
+            // Gets the properties for a task.
+            String type = fileInput.nextLine();
+            Boolean done = Boolean.parseBoolean(fileInput.nextLine());
```
hieunm1821: Should the name of this boolean variable be isDone?

hieunm1821 (commented):
Overall, I found you code easy to read. There is no code quality issues. I noted some coding standard related issues only. Good work. :+1:

src/main/java/duke/tasklist/TaskList.java
```diff
@@ -0,0 +1,133 @@
+package duke.tasklist;
```
yumengtan, line 1: Like how u organize the classes into respective packages! 👍

src/main/java/duke/storage/Storage.java
```diff
@@ -0,0 +1,80 @@
...
+import java.io.File;
+import java.io.FileNotFoundException;
+import java.io.FileWriter;
+import java.io.IOException;
+import java.util.Scanner;
+
+/**
+ * The Storage object acts as an interface between Duke and the filename for database management.
+ */
+public class Storage{
```
yumengtan: Perhaps there should be a " " separator before the {. I have noticed that in ListIndexOutOfBound class as well!

src/main/java/duke/storage/Storage.java
yumengtan: Perhaps using a prefix such as is to the variable name such as IsDone like @hieunm1821 mentioned would make it sound more like booleans!

yumengtan (commented):
Overall your code structure is very well organised! Just some nits to fix, well done! 👍

src/main/java/duke/dukeexceptions/DukeExceptions.java
Denniszedead: Duke exception is a specialised exception for duke so I actually prefer to name it that way.

src/main/java/duke/dukeexceptions/EmptyDate.java
Denniszedead: Yeap! You are right! I'll make the changes!

src/main/java/duke/storage/Storage.java
Denniszedead: Yeap! Must have overlooked it! Thanks for the notice!

src/main/java/duke/storage/Storage.java
Denniszedead: Thanks! Must have overlooked it!

PR #262, opened 2022-01-27
[Benjamin Koh Wei Jie] iP

Benjykoh:
# SQUISHYBOT

> "You're the bubble to my tea" - some guy on the internet

### Squishy will help you remember the things you have set out to accomplish! It is:
- Easy to pick up
- GUI Based
- Extremely fast

### To utilise Squishy, 
* Clone this [repo](https://github.com/Benjykoh/ip/)
* Run `Launcher.java` with any IDE
* Have fun!

### Launch Squishy here!
```ruby
public class Launcher {
}
```

### Squishy can:
- [X] Add tasks (todo, events, deadlines)
- [X] Delete them ⚔️ 
- [X] List them 📖 
- [X] Find them 🔍 

### **What are you waiting for? Squishy is already ready to be your companion!**


src/main/java/Parser.java
```diff
@@ -3,13 +3,19 @@
 public class Parser {
 
     public static void parse(String input, TaskList taskList) {
-        switch(input) {
+        String[] array = input.split(" ");
+        String execute = array[0];
+        switch(execute) {
             case("list"):
```
mradhyaf: `switch` and `case` statements should have the same level of indentation according to the [coding standard](https://se-education.org/guides/conventions/java/intermediate.html)

src/main/java/TaskList.java
```diff
@@ -1,21 +1,21 @@
...
+
+    public TaskList() {
+        this.taskList = new ArrayList<>();
+    }
+
+    public void add(String task) {
+        taskList.add(task);
+    }
+
+    public int size() {
```
mradhyaf: Perhaps a verb such as getSize would be better for the name of this method?

mradhyaf:
Almost LGTM, just a few nits to pick.

All the best for your iP!

src/main/java/Parser.java
```diff
@@ -0,0 +1,103 @@
+import java.util.ArrayList;
+
+public class Parser {
+
+    public static void parse(String input, TaskList taskList) {
+        String[] array = input.split(" ");
+        String execute = array[0];
+        switch(execute) {
+            case("list"):
```
Rye-Catcher: The switch statement should have the following form. There is no indentation for case clauses.

src/main/java/Printer.java
```diff
@@ -0,0 +1,86 @@
...
+    }
+
+    public static void event(String eventName, String eventTime, TaskList taskList) {
+        String temp = "\n" + BLANK_LINE + "     Got it. I've added this task: \n" + "       [E][ ]" + eventName +
+                "(at: " + eventTime + ")" + "\n" +
+                "     Now you have " + taskList.size() + " task(s) in the list.\n" + BLANK_LINE;
+        System.out.println(temp);
+    }
+
+    public static void delete(int numToDelete, TaskList taskList) {
```
Rye-Catcher: Since this method has a wide scope. Maybe it is better to have a name with a more direct meaning? Like `printTodo` e.t.c.

Rye-Catcher (commented):
Overall the code looks good. 👍  But many public methods are not commented using Java Doc style. Also, there are a few layout style violations.

PR #281, opened 2022-01-28
[Tan Gee Teng] iP

geetengtan:
DukePro frees your mind of having to remember things you need to do. It's
- text-based
- easy to learn

All you need to do is,
1. download it from here.
2. double-click it.
3. add your tasks.
4. let it manage your tasks for you 😉 👍 

And it is **FREE**

Features:
- [x] Managing tasks
- [x] Managing deadlines
- [ ] Reminders (coming soon)

If you Java programmer, you can use it to practice Java too. Here's the ```main``` method:
```
public class Main {
    }
}
```


src/main/java/duke/Deadline.java
```diff
@@ -0,0 +1,26 @@
...
+
+    @Override
+    public String toString() {
+        String tempStr = " ";
+
+        if (this.isDone) {
+            tempStr = "X";
+        }
+
+        return "[D][" + tempStr + "] " + this.description + "(" + this.time.get().format(dateFormatter) + ")";
```
yunbinmo: Not sure if this is a displaying problem, probably it would be better to put some line breaks to improve readability?

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,44 @@
+package duke;
+
+import java.util.Scanner;
+
+public class Duke {
```
yunbinmo, line 7: Probably some javadocs to describe the class would be good for future developers, this also applies to several other classes :)

src/main/java/duke/Event.java
```diff
@@ -0,0 +1,26 @@
...
+
+    @Override
+    public String toString() {
+        String tempStr = " ";
+
+        if (this.isDone) {
+            tempStr = "X";
+        }
+
+        return "[E][" + tempStr + "] " + this.description + "(" + this.time.get().format(dateFormatter) + ")";
```
yunbinmo: Similarly, perhaps it is better the add some line breaks to make the structure of the return message look clearer to reader.

src/main/java/duke/Task.java
```diff
@@ -0,0 +1,55 @@
+package duke;
+
+import java.util.Optional;
+import java.time.LocalDate;
+import java.time.format.DateTimeFormatter;
+
+abstract class Task {
+
+    protected static final DateTimeFormatter dateFormatter = DateTimeFormatter.ofPattern("LLL dd yyyy");
```
yunbinmo: Should the constant name be all uppercase using underscore to separate words?

src/main/java/duke/Ui.java
```diff
@@ -0,0 +1,73 @@
...
+        System.out.println("Hello from\n" + logo);
+
+        myPrint("Hello! I'm Duke\n    What can I do for you?");
+    }
+
+    public void exit() {
+        myPrint("Bye. Hope to see you again soon!");
+    }
+
+    public static void myPrint(String toPrint) {
```
yunbinmo: Should this method name be verbs?

yunbinmo (commented):
I like how you make most of the names sound straightforward and the code blocks are properly separated by empty lines!

src/main/java/duke/Deadline.java
geetengtan: yes it will!

src/main/java/duke/Duke.java
geetengtan: still not added yet sorry

src/main/java/duke/Task.java
geetengtan: you're right. will change it

src/main/java/seedu/duke/TaskList.java
```diff
@@ -85,4 +85,25 @@ public String count() {
         return "\n    Now you have " + tasks.size() + " task" + isSingular + " in your list.";
     }
 
+    public String find(String keyword) {
+        String resultStr = "";
+        boolean keywordFound = false;
```
seksek13: Boolean variable should use a prefix such as is, has, was.

src/main/java/seedu/duke/Ui.java
```diff
@@ -40,6 +44,9 @@ public void nextInput(String input, TaskList tasks) {
             case DELETE:
                 myPrint(tasks.delete(Parser.getIndex(input)));
                 break;
+            case FIND:
+                myPrint(tasks.find(Parser.getDescription(input)));
```
seksek13: Maybe it will be better to change the myprint() method name to something that would explain the function better

src/main/java/duke/TaskList.java
```diff
@@ -0,0 +1,149 @@
...
+
+    }
+
+    /**
+     * Returns a string that describes the number of tasks in the list
+     *
+     * @return String that describes the number of tasks in the TaskList
+     */
+    public String count() {
+        String isSingular = "s";
```
seksek13, line 119: Why does the string variable had the prefix "is", it might be confusing for others as this prefix is reserved for boolean variables. It might be better to rename this variable :)

src/test/java/duke/TaskListTest.java
```diff
@@ -0,0 +1,37 @@
...
+        assertEquals("Here are the tasks in your list:\n    1. [E][ ] party (Dec 12 2022)\n    2. [T][ ] borrow book", t.list());
+    }
+
+    @Test
+    void mark() {
+        ArrayList<Task> tasks = new ArrayList<Task>();
+        tasks.add(new Event("party ", LocalDate.of(2022, 12, 12)));
+        tasks.add(new ToDo("borrow book"));
+
+        TaskList t = new TaskList(tasks);
```
seksek13, line 29: is it possible to rename this variable name so that it would be clearer in showing what you are doing?

seksek13 (commented):
Overall, methods and variables had followed the naming convention, good job!

PR #294, opened 2022-01-29
[Edward Alvin] ip

ThePrevailingOne:


src/main/java/Duke.java
```diff
@@ -1,10 +1,76 @@
...
+            case EVENT:
+                wordList.storeEvent(value[0], value[1]);
+                break;
+            case BYE:
+                replyBye();
+                break;
+            case NONE:
+                break;
+        }
+    }
```
ckcherry23, line 63: The coding standard for this module requires no indentation for the `case` clauses under `switch`! Perhaps you can configure your IDE to follow this new standard by default.

src/main/java/InputType.java
```diff
@@ -0,0 +1,17 @@
+public enum InputType {
```
ckcherry23, line 1: Interesting way to use enums!

src/main/java/Deadline.java
```diff
@@ -0,0 +1,14 @@
+public class Deadline extends WordListItem{
```
ckcherry23, line 1:
```suggestion
public class Deadline extends WordListItem {
```
Be careful with the spacing requirements for Egyptian style brackets!

src/main/java/WordList.java
```diff
@@ -0,0 +1,67 @@
...
+        String str = "";
+        str += "------------------------------------\n";
+        for(WordListItem wordListItem: this.wordList) {
+            str += i + ". " + wordListItem + "\n";
+            i++;
+        }
+        str += "------------------------------------\n";
+        return str;
+    }
+}
```
ckcherry23, line 67: I think you can make use of the `toString()` method of `WordListItem` directly in `printList()`, instead of returning a complicated `String` expression here.

ckcherry23 (commented):
The code thus far looks good! 👍🏼
Just some suggestions and nits to fix.

PR #302, opened 2022-01-29
[Adrian Ong] iP

AdrianOngJJ:
### LeDuke
_FAST_, LITE, ~~POWERFUL~~ **MEGA STONKS**
LeDuke will help you with:
- Remembering tasks
- Keeping track of which tasks are completed

What? Did I hear that you want LeDuke? Follow these easy steps :eyes::

1. Find code [here](https://github.com/AdrianOngJJ/ip)
2. Run the JAR file
3. ...
4. PROFIT!!

Progress:

- [x] Mabaging tasks
- [ ] Managing deadlines
- [ ] Reminders

You can also use look for the `main` in my code:
```java
        new Duke("/tasks.txt").run();
    }
```


src/main/java/Duke.java
```diff
@@ -1,10 +1,135 @@
...
+     * @param bw BufferedWriter from main to print out the Master List.
+     * @throws java.io.IOException If an I/O error occurs. Only takes in string.
+     */
+    private static final void printList(BufferedWriter bw) throws Exception {
+        for(int i = 0; i < masterList.size(); i++) {
+            Task curr = masterList.get(i);
+            bw.write((i + 1) + "." + curr.toString());
+            bw.newLine();
+        }
+        bw.flush();
```
joshuayapwj98: A very intuitive set of code!

I was wondering if there's also a need for whitespace after the "for" word?

Nonetheless, neat work :)

src/main/java/DukeException.java
```diff
@@ -0,0 +1,11 @@
+/**
+ * Exception class for Duke to handle Exceptions
+ */
+
+public class DukeException extends Exception {
+
+    public DukeException(String message) {
+        super(message);
+    }
```
joshuayapwj98: Awesome that this class is kept simple!

However, maybe there should be some comments to document on what kind of exceptions you are expecting? 😄

src/main/java/Duke.java
```diff
@@ -1,10 +1,135 @@
...
                 + "|  _ \\ _   _| | _____ \n"
                 + "| | | | | | | |/ / _ \\\n"
                 + "| |_| | |_| |   <  __/\n"
                 + "|____/ \\__,_|_|\\_\\___|\n";
         System.out.println("Hello from\n" + logo);
+        printLineBreak();
+        System.out.println("Hello! I'm Duke\nWhat can I do for you?");
+        String input; // to store raw input command
+        String[] inputArr; // to store split input command
+        boolean ifBye = false;
```
joshuayapwj98: Intuitive boolean to check if the program should exit or not!

However, maybe:
```suggestion
boolean shouldAbort = false;
```

joshuayapwj98 (commented):
Overall, an elegant and tidy set of code! Can distinguish the flow of the programme and it would be easy to take over as a public repo if ever you are going to publish this project 😄 

Just some tiny pointers about the indentation and also whitespace! Otherwise, good job 👍🏼

src/main/java/Deadlines.java
```diff
@@ -0,0 +1,19 @@
+/**
+ * Deadlines tasks that need to be done before a specific date/time
+ * e.g., submit report by 11/10/2019 5pm
+ */
+
```
ovidharshini: Consider, perhaps, deleting this line break between the comments and the function for consistency

src/main/java/Duke.java
```diff
@@ -1,10 +1,135 @@
...
+ * @version 0.1
+ * @since 22/1/2022
+ */
 public class Duke {
-    public static void main(String[] args) {
+    private static final String LINE_BREAK = "\n~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~";
+    private static final ArrayList<Task> masterList = new ArrayList<>();
+    /**
+     * Prints line break.
+     * @return void
```
ovidharshini: Consider adding a line break between the description of a function and its associated Javadoc tags for consistency. I noticed the same issue across other functions as well.

ovidharshini: Should the @return tag here be deleted? It is not required here as per Oracle guidelines since the return type is void. Do refer to [Javadoc guidelines](https://www.oracle.com/sg/technical-resources/articles/java/javadoc-tool.html) for a more thorough reference for the conventions with respect to Javadoc comments.

src/main/java/Duke.java
```diff
@@ -1,10 +1,135 @@
...
+    private static final void printLineBreak() {
+        System.out.println(LINE_BREAK);
+    }
+
+    /**
+     * Prints Master List
+     * @param bw BufferedWriter from main to print out the Master List.
+     * @throws java.io.IOException If an I/O error occurs. Only takes in string.
+     */
+    private static final void printList(BufferedWriter bw) throws Exception {
```
ovidharshini: Nice job on catching this! Do consider throwing an explicit IOException if you mention it as such in the documentation for the code though

src/main/java/Duke.java
```diff
@@ -1,10 +1,135 @@
...
+        for(int i = 0; i < masterList.size(); i++) {
```
ovidharshini: You might want to include a whitespace after the 'for' keyword here?

src/main/java/Duke.java
```diff
@@ -1,10 +1,135 @@
...
+            Task curr = masterList.get(i);
+            bw.write((i + 1) + "." + curr.toString());
+            bw.newLine();
+        }
+        bw.flush();
+    }
+
+    private static final String getDateTime(String[] inputArr) {
+        return inputArr[1].split("/")[1].split(" ", 2)[1]; // split input from slash
```
ovidharshini: Perhaps include the comment as a same-level indent above the line as opposed to an inline one? I noticed this in other places as well. 

Since this function consists of a single line at the moment, you might want to add Javadoc-style documentation instead.

src/main/java/Duke.java
```diff
@@ -1,10 +1,135 @@
...
+        String input; // to store raw input command
+        String[] inputArr; // to store split input command
+        boolean ifBye = false;
+        do {
+            printLineBreak();
+            System.out.println();
+            input = br.readLine();
+            inputArr = input.split(" ", 2); // split first word from body
+            printLineBreak();
+            switch (inputArr[0]) {
```
ovidharshini: Do consider separating the code into logical blocks with line breaks for ease of viewing. For instance, here, the do-while loop can be separated from the rest of the preceding lines, and the initialization of variables (e.g. BufferedReader) can be separated in a similar fashion as well.

src/main/java/Duke.java
```diff
@@ -1,10 +1,135 @@
...
+            case "bye":
+                ifBye = true;
+                System.out.println("Bye. Hope to see you again soon!");
+                break;
+
+            case "list":
```
ovidharshini: Good job on following the guidelines with respect to the indentation levels of switch and case! There should be no need for line breaks between individual case statements, so maybe remove these as a final step?

ovidharshini (commented):
LGTM after sorting out a few nits. Overall, the code adhered well to Java coding conventions. Spacing of comments and code appears to be irregular at times, so do keep a look out for them. You can also consider adding file types of ".class" to gitignore so as to not push them to remote. 

Good job!

PR #303, opened 2022-01-30
[Han Jiyao] iP

HanJiyao:
# Duke Chat

- text-based
- easy to learn

All you need to do is,

1. download it from [here](https://github.com/nus-cs2103-AY2122S2/ip.git).
2. double-click it.
3. add your tasks.

And it is **FREE!**

### Features:

- [x] Managing tasks
- [x] Managing deadlines 
- [x] Reminders (coming soon)

---

```java
public class Main {
    }
}
```


src/main/java/Deadline.java
```diff
@@ -0,0 +1,26 @@
...
+
+    @Override
+    public Deadline unmark() {
+        return new Deadline(task, false);
+    }
+
+    @Override
+    public String toString() {
+        return "["+Type.D+"]" + super.toString() + " (by: " + by + ")";
+    }
```
hsiaojietng: Well done on the naming of variables, they are easily read!

src/main/java/Duke.java
```diff
@@ -1,10 +1,146 @@
...
+                            }
+                            list.add(new ToDo(todo.toString()));
+                            System.out.println(output("Got it. I've added this task: \n        "
+                                    + list.get(list.size() - 1) + "\n    Now you have " + list.size() + " tasks in the list."));
+                        }
+                        break;
+
+                    case "deadline":
+                        StringBuilder deadline = new StringBuilder();
+                        StringBuilder deadlineBy = new StringBuilder();
```
hsiaojietng: Maybe can encapsulate this functionality in another class if you want

src/main/java/Event.java
```diff
@@ -0,0 +1,26 @@
+public class Event extends Task{
+    protected String at;
+    public Event(String task, String at) {
+        super(task.trim());
+        this.at = at.trim();
+    }
+
+    public Event(String task, boolean done) {
```
hsiaojietng: Easy to read!

hsiaojietng (commented):
LGTM. If you're able to encapsulate the behaviour of all the actions of Duke then it would be better

src/main/java/command/AddDeadlineCommand.java
```diff
@@ -0,0 +1,24 @@
...
+public class AddDeadlineCommand extends Command {
+    private final task.Deadline newDeadline;
+
+    public AddDeadlineCommand(Deadline newDeadline) {
+        super();
+        this.newDeadline = newDeadline;
+    }
+
+    public void execute(TaskList tasks, Ui ui, Storage storage) {
+        tasks.add(new Deadline(newDeadline.getTask(), newDeadline.getDate()));
```
9teMare: I like how you use getter method to keep fields in `Deadline` class private. Good job 👍!

src/main/java/command/AddDeadlineCommand.java
```diff
@@ -0,0 +1,24 @@
...
+        storage.saveTaskList(tasks);
+        ui.showMessage("Got it. I've added this task: \n        "
```
9teMare: Extra blank spaces after the new line character is a little bit misleading, maybe can move these blank spaces into the next line.

src/main/java/command/AddDeadlineCommand.java
```diff
@@ -0,0 +1,24 @@
...
+                + tasks.getByIndex(tasks.getSize() - 1) + "\n    Now you have " + tasks.getSize() + " tasks in the list.");
```
9teMare: Not sure if this line of code is too long for our coding standard, but generally try to break up a long line of code like this.

src/main/java/command/ErrorCommand.java
```diff
@@ -0,0 +1,18 @@
+package command;
+
+import storage.Storage;
+import task.TaskList;
+import ui.Ui;
+
+public class ErrorCommand extends Command {
```
9teMare: I love your implementation of having an abstract class `Command` and various other commands that extend it, however, personally I do not think an `ErrorCommand` is considered as a `Command`, because if you think about it, it is not like the user wants to input something intentionally that triggers this. The code works perfectly fine, it's just that why not throw exception directly when parsing the user input in the first place, instead of having recognized that the user input something invalid but still parse it to a `Command`?

src/main/java/command/ExitCommand.java
```diff
@@ -0,0 +1,19 @@
...
+        super();
+    }
+
+    @Override public void execute(TaskList tasks, Ui ui, Storage storage) {
+        ui.showExitMessage();
+    }
+
+    @Override public boolean isExit() {
+        return true;
+    }
```
9teMare: The `@Override` tag should occupy its own line.

src/main/java/command/ListCommand.java
```diff
@@ -0,0 +1,21 @@
...
+public class ListCommand extends Command {
+
+    @Override
+    public void execute(TaskList tasks, Ui ui, Storage storage) {
+        StringBuilder text = new StringBuilder();
+        text.append("Here are the tasks in your list:\n");
+        for (int i = 0; i < tasks.getSize(); i++) {
+            text.append("    ").append(i + 1).append(". ")
+                    .append(tasks.getByIndex(i))
+                    .append("\n");
```
9teMare: I do not think the indentation here follows the coding standard ☹️.

src/main/java/command/SearchCommand.java
```diff
@@ -0,0 +1,32 @@
...
+
+    @Override
+    public void execute(TaskList tasks, Ui ui, Storage storage) {
+        StringBuilder searchText = new StringBuilder();
+        searchText.append("Here are the tasks on ").append(date.format(DateTimeFormatter.ofPattern("MMM dd yyyy"))).append(" \n");
+        for (int i = 0; i < tasks.getSize(); i++) {
+            if (tasks.getByIndex(i).getDate() != null && tasks.getByIndex(i).getDate().equals(date)) {
+                searchText.append("    ").append(i + 1).append(". ")
+                        .append(tasks.getByIndex(i))
+                        .append("\n");
```
9teMare: Same here

src/main/java/parser/Parser.java
```diff
@@ -0,0 +1,131 @@
+package parser;
+
+import command.AddDeadlineCommand;
+import command.Command;
+import command.ExitCommand;
+import command.MarkCommand;
+import command.ListCommand;
+import data.DukeException;
```
9teMare: Not sure if this line follows the import order in checkStyle.

src/main/java/storage/Storage.java
```diff
@@ -0,0 +1,89 @@
...
+                switch (input.get(0)) {
+                    case "T":
+                        ToDo newTodo = new ToDo(input.get(2).trim());
+                        list.add(newTodo);
+                        if (input.get(1).equals("1")) {
+                            list.set(list.indexOf(newTodo), newTodo.mark());
+                        }
+                        break;
+                    case "D":
+                        Deadline newDeadline = new Deadline(input.get(2).trim(), LocalDate.parse(input.get(3).trim(), formatter));
```
9teMare: Sufficient comment here can probably improve the readability by a lot.

src/main/java/task/Deadline.java
```diff
@@ -0,0 +1,42 @@
+package task;
+
+import java.time.LocalDate;
+import java.time.format.DateTimeFormatter;
+
+public class Deadline extends Task {
+    protected LocalDate by;
```
9teMare: `private` should be a better accesser here than `protected` because there are no other classes extending from Deadline, therefore, the only object that will access this field is of type `Deadline`. Declaring this field as `final` should also be a good practice because the time should not be changed after `new()` an object.

src/main/java/task/Deadline.java
```diff
@@ -0,0 +1,42 @@
...
+    }
+
+    @Override
+    public Deadline unmark() {
+        return new Deadline(task, by, false);
+    }
+
+    @Override
+    public String saveData() {
+        int done = super.done? 1 : 0;
```
9teMare: Declaring this as `boolean isDone` might be better.

src/main/java/task/Task.java
```diff
@@ -0,0 +1,43 @@
...
+        this.done = done;
+    }
+
+    public Task mark() {
+        return new Task(task, true);
+    }
+
+    public Task unmark() {
+        return new Task(task, false);
+    }
```
9teMare: Not too sure about the rationale behind the need to `new()` an object, (same for `Deadline` `Event` and `Todo`), why not just simply set the `done` field as true or false, all the children of `Task` share the same behaviour and they can just access `mark()` and `unmark()` straight away.

9teMare (commented):
Overall good job 🙂! 
Just a few points to mention:
- OOP practices can be implemented better
- Coding standard should be followed consistently
- Seems like you are a little bit behind schedule, looking forward to your GUI implementation!

PR #305, opened 2022-01-30
[Edward Alvin] ip

ThePrevailingOne:


src/main/java/Duke.java
```diff
@@ -1,10 +1,36 @@
...
         System.out.println("Hello from\n" + logo);
+        String input = sc.nextLine();
+        while (!isBye(input)) {
+            echo(input);
+            input = sc.nextLine();
+        }
+        replyBye();
+    }
+
+    public static void echo(String words) {
```
geraldneo567: Perhaps a more descriptive method name like `printLines` would be clearer!

src/main/java/Duke.java
```diff
@@ -26,54 +22,36 @@ public static void main(String[] args) {
...
 
-    public static boolean isUnmark(String word) {
-        return word.startsWith(Duke.UNMARK);
-    }
-
-    public static boolean isList(String word) {
-        return Duke.LIST.equals(word);
+    public static void processInput(InputType inputType, String value) {
+        switch(inputType) {
+            case LIST:
```
geraldneo567: According to the codestyle, there should be no intentation for 'case' clauses in the switch block.

src/main/java/Duke.java
```diff
@@ -26,54 +22,36 @@ public static void main(String[] args) {
...
     }
 
-    public static boolean isUnmark(String word) {
-        return word.startsWith(Duke.UNMARK);
-    }
-
-    public static boolean isList(String word) {
-        return Duke.LIST.equals(word);
+    public static void processInput(InputType inputType, String value) {
+        switch(inputType) {
```
geraldneo567: Minor style-related thing, but there should be a space before the bracket i.e `switch (inputType)`

src/main/java/Duke.java
```diff
@@ -25,7 +25,7 @@ public static void main(String[] args) {
 
             Object[] parseResult = InputParser.parseInput(input);
             InputType inputType = (InputType) parseResult[0];
-            String value = (String) parseResult[1];
+            String[] value = (String[]) parseResult[1];
```
geraldneo567: Why the cast to String[] array here, instead of making `InputParser::parseInput(input)` return a String[] instead of Object[]?

src/main/java/InputParser.java
```diff
@@ -1,13 +1,31 @@
...
         for(InputType inputType: InputType.values()) {
+            if (inputType == InputType.NONE) {
+                continue;
+            }
+
             if (input.startsWith(inputType.label)) {
-                value = "";
-                if (inputType.label.length() < input.length()) {
-                    value = input.substring(inputType.label.length() + 1);
+                if (inputType == InputType.BYE || inputType == InputType.LIST) {
```
geraldneo567: Maybe it would be neater to separate out into boolean variables for example `isBye`, `isList`, etc to be in line with the code quality guideline

src/main/java/Duke.java
```diff
@@ -1,10 +1,85 @@
...
+
+        replyWelcomeMessage();
+        String input;
+        while (true) {
+            try {
+                input = sc.nextLine();
+                checkEmpty(input);
+                Object[] parseResult = InputParser.parseInput(input);
+                InputType inputType = (InputType) parseResult[0];
+                String[] value = (String[]) parseResult[1];
```
muraddurrani: I think parseResult and value could both be renamed in plural form (results and values perhaps?) to be consistent with naming conventions. I noticed this in a few other places.

src/main/java/Duke.java
```diff
@@ -1,10 +1,85 @@
...
+                System.out.println(e);
+            } catch (NumberFormatException e) {
+                System.out.println(e);
+            } catch (Exception e) {
+                System.out.println(e);
+            }
+        }
+    }
+
+    public static void processInput(InputType inputType, String[] value) {
```
muraddurrani: Small thing because its just a parameter but maybe change value to values so its plural?

src/main/java/Todo.java
```diff
@@ -0,0 +1,12 @@
+public class Todo extends WordListItem{
```
muraddurrani: I think there should be a space between WordListItem and {.

src/main/java/UnknownInputException.java
```diff
@@ -0,0 +1,5 @@
+public class UnknownInputException extends Exception{
```
muraddurrani: I think there should be a space between Exception and {.

src/main/java/WordList.java
```diff
@@ -0,0 +1,67 @@
+import java.util.ArrayList;
+
+public class WordList {
+    private ArrayList<WordListItem> wordList;
```
muraddurrani: Perhaps this could be renamed to words or something similar to be plural?

src/main/java/WordList.java
```diff
@@ -0,0 +1,67 @@
...
+        this.wordList.get(itemNumber - 1).unmarkItem();
+        System.out.println("Nice! I've marked this task as not done: ");
+        System.out.println("  " + this.wordList.get(itemNumber - 1));
+    }
+
+    public void printList() {
+        System.out.println(this);
+    }
+
+    public int length() {
```
muraddurrani: The coding standards suggest that method names be verbs. Perhaps this could be changed to getLength or something like that.

muraddurrani (commented):
Generally I found it quite easy to read all the code. I think you just need to adjust some small things like renaming arrays and arraylists to be plural form.

ThePrevailingOne:
# Duked
> “Your mind is for having ideas, not holding them.” – David Allen


Duked frees your mind of having to remember things you need to do. It's,
-   text-based
-   easy to learn
-   ~~FAST~~ _SUPER_ FAST to use

All you need to do is,
1.  download it from [here](https://nus-cs2103-ay2122s2.github.io/website/admin/ip-w4.html).
2.  double-click it.
3.  add your tasks.
4.  let it manage your tasks for you 😉

And it is **FREE**!

Features:
- [x]    Managing tasks
- [x]   Managing deadlines (coming soon)
- [x]   Sorting tasks (coming soon)

Just double click on the jar and it should open the app GUI :)

or

Run 
```bash
java -jar Duke.jar //depending on the relative location of the app
```

PR #309, opened 2022-01-31
[Teh Kok Hoe] iP

tehkokhoe:


src/main/java/Duke.java
```diff
@@ -1,10 +1,166 @@
+import java.util.*;
+
 public class Duke {
+    enum Command {
```
bakano98: Good to see you're using enums for this! Keep it up.

src/main/java/Task.java
```diff
@@ -0,0 +1,50 @@
+public class Task {
+    private String task;
+    private boolean done;
```
bakano98: According to the coding standards, this should be named `isDone`

src/main/java/Duke.java
```diff
@@ -1,10 +1,166 @@
...
                 + "|  _ \\ _   _| | _____ \n"
                 + "| | | | | | | |/ / _ \\\n"
                 + "| |_| | |_| |   <  __/\n"
                 + "|____/ \\__,_|_|\\_\\___|\n";
         System.out.println("Hello from\n" + logo);
+        System.out.println("    ____________________________________________________________");
+        System.out.println("    Hello! I'm Duke");
+        System.out.println("    What can I do for you?");
+        System.out.println("    ____________________________________________________________");
+        boolean StillIn = true;
```
bakano98: Use camelCase to name variables, and booleans should be of the form `isSomething`. You could possibly rename this to `isStillIn`

src/main/java/Duke.java
```diff
@@ -1,10 +1,166 @@
...
+        FastReader fr = new FastReader();
+        ArrayList<Task> list = new ArrayList<Task>(100);
+        while(StillIn) {
+            String in = fr.nextLine();
+            String[] input = in.split(" ", 2);
+            Command cmd = Command.valueOf(input[0].toUpperCase());
+            System.out.println("    ____________________________________________________________");
+            try {
+                switch (cmd) {
+                    case BYE:
```
bakano98: `case` should not be indented. It should be on the same line as `switch`

src/main/java/Duke.java
```diff
@@ -1,10 +1,166 @@
...
+    }
+
     public static void main(String[] args) {
         String logo = " ____        _        \n"
                 + "|  _ \\ _   _| | _____ \n"
                 + "| | | | | | | |/ / _ \\\n"
                 + "| |_| | |_| |   <  __/\n"
                 + "|____/ \\__,_|_|\\_\\___|\n";
         System.out.println("Hello from\n" + logo);
+        System.out.println("    ____________________________________________________________");
```
bakano98: Instead of using 4 spaces, you can consider using `\t` instead for tab

src/main/java/Duke.java
```diff
@@ -1,10 +1,166 @@
+import java.util.*;
```
bakano98: Two things here: 
1. Keep everything in a package
2. Only import things that are needed to be used. Avoid using wildcard imports

src/main/java/FastReader.java
```diff
@@ -0,0 +1,51 @@
+import java.io.BufferedReader;
```
bakano98: Not too sure if `FastReader` is required here, but good consideration!

src/main/java/Duke.java
```diff
@@ -1,10 +1,166 @@
...
                 + "|____/ \\__,_|_|\\_\\___|\n";
         System.out.println("Hello from\n" + logo);
+        System.out.println("    ____________________________________________________________");
+        System.out.println("    Hello! I'm Duke");
+        System.out.println("    What can I do for you?");
+        System.out.println("    ____________________________________________________________");
+        boolean StillIn = true;
+        FastReader fr = new FastReader();
+        ArrayList<Task> list = new ArrayList<Task>(100);
+        while(StillIn) {
```
bakano98: I think you need to put spacing between the brackets and each `while` and each `if`. You only do not use spaces when it is a function/method declaration

bakano98 (commented):
A few coding standard violations, but other than that, nothing too huge. Remember to put everything into a package.
LGTM after fixing these.

PR #316, opened 2022-02-01
[Sim Jia Ming] iP

SimJM:
 
 * text-based
 * easy to learn
 * ~FAST~ _SUPER_ FAST to use
   All you need to do is,
 
 2. double-click it.
 3. add your tasks.
 4. let it manage your tasks for you 💯
 
 its **FREE**
 
 tasklist:
 
 * [x]  done
 * [ ]  not done
 
 Here's the main method:
 ```
 public class Main {
    }
}
 ```


src/main/java/Deadline.java
```diff
@@ -0,0 +1,20 @@
...
+    protected final String sym = "D";
+    protected String day;
+
+    Deadline (String description, String day) {
+        super(description);
+        this.day = day;
+    }
+
+    @Override
+    String getSym() {
```
tyanhan: I would opt for a more intuitive name rather than sym.

src/main/java/Duke.java
```diff
@@ -1,10 +1,107 @@
...
+        String DASH = "____________________________________________________________";
+        Action act = new Action();
+
+        act.greet();
+
+        while (sc.hasNext()) {
+            String phrase = sc.nextLine();
+            System.out.println(DASH);
+
+            if (phrase.equals("list")) {
```
tyanhan: Perhaps you can consider using a switch statement instead?

src/main/java/Duke.java
```diff
@@ -1,10 +1,107 @@
...
+                            }
+                        }
+                        Event t = new Event(remainingWords, dayAndTime);
+                        arrlst.add(new Task(t.description, t.sym));
+                        System.out.println(gotIt);
+                        System.out.println(t);
+                        String noOfTask = String.format("Now you have %d tasks in the list.", arrlst.size());
+                        System.out.println(noOfTask);
+                    }
+                } catch (ArrayIndexOutOfBoundsException aioobe) { // echo
```
tyanhan: You can consider changing the parameter name to just ```e``` instead.

src/main/java/Duke.java
tyanhan: The code here is pretty long, but you will get to break it down later as you progress with the iP. Consider doing the printing in separate functions and get your switch statement to invoke those print functions instead.

tyanhan (commented):
Good job overall, I did not notice any basic styling errors. You can improve your code by breaking down long parts into separate functions. All the best!

src/main/java/Task.java
```diff
@@ -0,0 +1,45 @@
+public class Task {
```
yongler: It is possible to change Task to an abstract class?

src/main/java/Action.java
```diff
@@ -0,0 +1,35 @@
+import java.util.ArrayList;
+
+public class Action {
```
yongler: Looks clean!

src/main/java/Deadline.java
yongler: > I would opt for a more intuitive name rather than sym.

Maybe you can change it to a more descriptive name such as symbol?

src/main/java/Duke.java
```diff
@@ -1,10 +1,107 @@
...
     public static void main(String[] args) {
-        String logo = " ____        _        \n"
-                + "|  _ \\ _   _| | _____ \n"
-                + "| | | | | | | |/ / _ \\\n"
-                + "| |_| | |_| |   <  __/\n"
-                + "|____/ \\__,_|_|\\_\\___|\n";
-        System.out.println("Hello from\n" + logo);
+        Scanner sc = new Scanner(System.in);
+        ArrayList<Task> arrlst = new ArrayList<>();
+        String DASH = "____________________________________________________________";
```
yongler: This works too, but UI can be separated into a class for cleanliess

yongler (approved):
LGTM! Just some minor nits to fix

src/main/java/Task.java
```diff
@@ -0,0 +1,45 @@
...
+    }
+
+    String getSym() {
+        return this.sym;
+    }
+
+    @Override
+    public String toString() {
+        return String.format("added: %s", description);
+    }
```
jetrz: Could use some javadocs documentation as Task is the parent class.

src/main/java/Duke.java
```diff
@@ -1,10 +1,107 @@
...
+                        for (int i = 1; i < arrWords.length; i++) {
+                            if (arrWords[i].equals("/at")) {
+                                for (int j = i + 1; j < arrWords.length; j++) {
+                                    dayAndTime = dayAndTime + " " + arrWords[j];
+                                }
+                                break;
+                            } else {
+                                remainingWords = remainingWords + " " + arrWords[i];
+                            }
+                        }
```
jetrz: Could use comments in places like these to explain what the code is doing (i.e. splitting up the user input) as it may not be apparent on first sight.

src/main/java/Action.java
```diff
@@ -0,0 +1,35 @@
...
+    }
+
+    void showList(ArrayList<Task> arrlst) {
+        System.out.println("Here are the tasks in your list:");
+        for (int i =  0; i < arrlst.size(); i++) {
+            String output = String.format("%d.[%s][%s]%s\n", i + 1, arrlst.get(i).sym,
+                                             arrlst.get(i).getStatusIcon(), arrlst.get(i).description);
+            System.out.println(output);
+        }
+    }
```
jetrz: Could use a javadocs documentation.

src/main/java/Event.java
```diff
@@ -0,0 +1,21 @@
...
+
+    Event (String description, String dayAndTime) {
+        super(description);
+        this.dayAndTime = dayAndTime;
+    }
+
+    @Override
+    String getSym() {
+        return this.sym;
+    }
```
jetrz: May I check what is the purpose of this override? Same with deadline and todo classes. Seems redundant as the getSym() in Task is the same.

jetrz (commented):
Generally clean code, well documented, styled, and easy to understand. Only minor nitpicks in terms of documentation and code logic here and there. Well done!

src/main/java/Task.java
```diff
@@ -0,0 +1,45 @@
+public class Task {
+    protected String description;
```
KwanHW: While protected may appear sufficient, these attributes can still be accessible by other functions once you start putting them into packages. Consider changing your attributes to `private`

src/main/java/Task.java
```diff
@@ -0,0 +1,45 @@
...
+        System.out.println(output);
+    }
+
+    String getSym() {
+        return this.sym;
+    }
+
+    @Override
+    public String toString() {
+        return String.format("added: %s", description);
```
KwanHW: I think you should change your toString() method to print out the `sym`, `isDone` and `description` instead of this message. This can reduce code duplication in your children classes as I observed that they are using attributes in `Task` to form the string.

src/main/java/Task.java
```diff
@@ -0,0 +1,45 @@
...
+    }
+
+    public void markAsNotDone() {
+        this.isDone = false;
+        System.out.println("OK, I've marked this task as not done yet:");
+        String output = String.format("  [%s] %s", this.getStatusIcon(), this.description);
+        System.out.println(output);
+    }
+
+    String getSym() {
```
KwanHW: Missing `public` keyword at function signature

KwanHW:
Good effort overall.

Some improvements to consider:
- Possible use of enums for `sym`
- Naming of variables (as mentioned by your classmates)
- [Code duplication / Abstraction Task.toString()](https://github.com/nus-cs2103-AY2122S2/ip/pull/316#discussion_r798825551)
- Double check your [function signature headers](https://github.com/nus-cs2103-AY2122S2/ip/pull/316#discussion_r798829927) throughout your project

